PR #3, opened 2017-04-11
Label Value View Implementation

emilyrohrbough:
### Summary
Implemented the react component LabelValueView to replace TitleValueCard. 

### Additional Details
The upgrade to React 15.5 is non-passive. So I locked react down at 15.4.2, as done in terra-core in [this pull request](https://github.com/cerner/terra-core/pull/202).


packages/terra-clinical-label-value-view/lib/LabelValueView.scss
```diff
@@ -0,0 +1,17 @@
...
+
+.terraClinical-LabelValueView {
+  border: 0;
+  padding: 0.1785rem;
+}
+
+.terraClinical-LabelValueView-label {
+  color: $terra-grey-50;
+  font-size: 1rem;
+  line-height: 20px;
```
StephenEsser: Recently we had a discussion to keep line-heights unit-less (dropping px / rem): https://css-tricks.com/almanac/properties/l/line-height/#article-header-id-0 

This line-height value is also different than the one used below

packages/terra-clinical-label-value-view/lib/LabelValueView.scss
```diff
@@ -0,0 +1,17 @@
...
+}
+
+.terraClinical-LabelValueView-value {
+  font-size: 1.4285714285714286rem;
+  line-height: 1.4285714285714286rem;
```
StephenEsser: Dave brought up a good point about this styling. If a String is passed in as a value we should style it based on UX feedback, but if a component was passed in, we probably shouldn't apply a font-size and line-height.

packages/terra-clinical-label-value-view/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "terra-clinical-label-value-view",
+  "main": "lib/LabelValueView.js",
+  "private": true,
+  "version": "0.0.0",
+  "description": "terra-clinical-label-value-view",
```
StephenEsser: A description should be added here, most people copy paste the description from the readme

packages/terra-clinical-label-value-view/lib/LabelValueView.scss
emilyrohrbough: The current implementation uses this UX styling based on the styling sheet for the TitleValueView. As for any children (say <h1>) or components that are passed in, these styling are override. What is a better way to set these values?

packages/terra-clinical-label-value-view/package.json
emilyrohrbough: [Added here](https://github.com/cerner/terra-clinical/pull/3/commits/5e8c8240b0286046986ffcd332075f19157fafb3#diff-fa0a9e8fe1159067a6842a6e15f71691R6)

packages/terra-clinical-label-value-view/lib/LabelValueView.scss
emilyrohrbough: [Fixed here](https://github.com/cerner/terra-clinical/pull/3/commits/5e8c8240b0286046986ffcd332075f19157fafb3#diff-700c50a3087237b9354196994a1d84f2R11)

packages/terra-clinical-label-value-view/lib/LabelValueView.scss
ryanthemanuel: Could we dynamically set this class in the jsx file based on what children is?  If it's a solitary string we add the class, otherwise not.  I'm not 100% sold on this idea so I'd like other people to weigh in as well.

tbiethman: We could do two separate props instead of children. `valueText` and `valueElement` or `valueComponent` or something. I'd (personally) like to avoid type-checking props. Or we could still use children too, something like:

```
<div {...customProps} className={labelValueViewClassNames}>
  <div className="terraClinical-LabelValueView-label">{props.label}</div>
  <div className="terraClinical-LabelValueView-value">{props.valueText}</div>
  {children}
</div>
```
We'd just need to document that valueText will be styled a certain way, and that children could be used for more substantial components.

ryanthemanuel: I'm cool with Tyler's suggestion.  I definitely would like to see this addressed before it gets merged in.

StephenEsser: +1 To Tyler's suggestion of having a prop for value but keeping children, I would conditionally render the value wrapper.

The Button does something similar: https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L73

When trying to render null nothing happens.

packages/terra-clinical-label-value-view/src/LabelValueView.jsx
```diff
@@ -0,0 +1,51 @@
...
+   */
+  label: PropTypes.string.isRequired,
+  /**
+  * The text to be displayed underneath the label with the provided styling.
+  */
+  textValue: PropTypes.string,
+  /**
+   *  Child component(s) to display underneath the label.
+   */
+  children: PropTypes.element,
```
StephenEsser: I would keep children as type node

emilyrohrbough: Yeah, I wasn't sure when I changed this. [Changed Back Here](https://github.com/cerner/terra-clinical/pull/3/commits/be4eaacb32bf32793566aebbf851eef3179064ea#diff-00cdf59bc370c025c9c00e3dbd5c61d0R41)

Issue #29, opened 2017-05-02
Don't place the accessory divs on the dom if not necessary to display.


dkasper-was-taken:
### Description of Issue
The accessory element should not be added to the dom if an accessory isn't present, the renderAccessory function should return undefined in the case that the accessory isn't provided, rather than a blank div.

PR #30, opened 2017-05-02
Do not add flex item for accessory section is the accessory isn't pre…

dkasper-was-taken:
Summary:
Do not add the empty div for accessories if the accessory element is not present. This prevents the unneeded margin from being added to the item view and throwing off alignment.

Addresses #29

### DIFF
--- a/packages/terra-clinical-item-view/src/ItemView.jsx
+++ b/packages/terra-clinical-item-view/src/ItemView.jsx
@@ -53,11 +53,16 @@ const defaultProps = {
 
 class ItemView extends React.Component {
 
-  static renderAccessory(accessory) {
+  static renderAccessory(accessory, accessoryAlignment) {
+    const accessoryClassNames = classNames([
+      'terraClinical-ItemView-accessory',
+      { [`terraClinical-ItemView-accessory--${accessoryAlignment}`]: accessoryAlignment },
+    ]);
+
     let accessorySection;
     if (accessory) {
       accessorySection = (
-        <div className="terraClinical-ItemView-accessory">
+        <div className={accessoryClassNames}>
           {accessory}
         </div>
       );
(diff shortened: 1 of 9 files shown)


packages/terra-clinical-item-view/tests/jest/__snapshots__/ItemView.test.jsx.snap
```diff
@@ -2,8 +2,6 @@ exports[`test should render 1 start theme display 1`] = `
 <div
   className="terraClinical-ItemView terraClinical-ItemView--oneColumn terraClinical-ItemView-accessory--alignCenter">
```
emilyrohrbough: `terraClinical-ItemView-accessory--alignCenter` is still assigned to the className even when no accessory is present. Should this happen?

dkasper-was-taken: I guess it's not necessary, but the intention was to leave it there since it's more a style setting on the item-view, that could potentially be moved in the code to sit on the accessory.

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -154,18 +163,17 @@ class ItemView extends React.Component {
...
       { 'terraClinical-ItemView--isTruncated': isTruncated },
       { [`terraClinical-ItemView--${layout}`]: layout },
-      { [`terraClinical-ItemView-accessory--${accessoryAlignment}`]: accessoryAlignment },
       customProps.className,
     ]);
 
     return (
       <div {...customProps} className={viewClassNames}>
-        {ItemView.renderAccessory(startAccessory)}
+        {ItemView.renderAccessory(startAccessory, accessoryAlignment)}
```
StephenEsser: I think you can conditionally render this inline with `startAccessory && ItemView.renderAccessory(startAccessory)` but what you have seems to work just fine too

Issue #40, opened 2017-05-15
Update the error view to use a more proper svg


ryanthemanuel:
Please fill out the below template as best you can.
--------------------------------------------------------

### Description of Issue
We need to update the image in the error view to use a more proper svg.  There was one that was used in legacy terra rails view models that we should use until the glyphs are more officially released.

PR #43, opened 2017-05-15
Change error icon to the svg used in terra-legacy

emilyrohrbough:
### Summary
Update the image in the error view to be the svg used in terra-legacy. Addresses issue #40. 

@cerner/terra


packages/terra-clinical-error-view/src/ErrorIcon.jsx
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable */
+import React from 'react';
+
+const fillColor = '#e8e9ea';
+
+const ErrorIcon = () => {
+  return (
+    <svg id="Layer_1" data-name="Layer 1" viewBox="0 0 251.7 251.8" style={{ height: '170px', width: '170px' }}>
```
Matt-Butler: nit: you can remove style and add height and width attributes

PR #51, opened 2017-05-17
Bump Dependency Versions

emilyrohrbough:
### Summary
This PR does two things:
1. Removes the 15.4.2 dependency lock
2. Bumps the dependency and peer dependency versions of terra components to the most recent release. This is needed to ensure 15.5 compatibility for installed dependencies.

@cerner/terra


packages/terra-clinical-action-header/package.json
```diff
@@ -23,29 +23,29 @@
...
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^0.4.1",
+    "terra-button": "^0.4.1",
+    "terra-button-group": "^0.2.1",
+    "terra-icon": "^0.4.1",
+    "terra-mixins": "^1.1.1",
     "terra-clinical-header": "^0.x",
-    "terra-responsive-element": "^0.x"
+    "terra-responsive-element": "^0.3.1"
```
Matt-Butler: You might already know this, but just a heads up.

"^0.3.1" will not resolve "0.x", 

it will only resolve to "0.3.x"

Matt-Butler: If you want to set a minimum version you could do something like this:
```
0.x > 0.3.1
```
screen shot 2017-05-17 at 4 05 10 pm

emilyrohrbough: Looking at this now, I see that. Terra-core did the versioning with the `^0.3.1` syntax. Should this be changed? Or would it be better to close this request & just remove the react dependency lock?

Matt-Butler: I would just update

```
"react": "15.4.2",
"react-dom": "15.4.2",
```
to
```
"react": "^15.4.2",
"react-dom": "^15.4.2",
```

If you leave the dependencies from terra-core at `0.x` then you will get new changes as we release them.  We just need to be careful in terra-core about preventing/communicating non-passive changes.

Matt-Butler: Let's post scrum about this tomorrow

emilyrohrbough:
Closing the PR to keep dependencies & peer dependencies at "^0.x".

PR #52, opened 2017-05-18
Remove the 15.4.2 dependency lock

emilyrohrbough:
### Summary
Removes the 15.4.2 dependency lock

@cerner/terra


mjhenkes:
Should we bump to react 15.5 instead of 15.4.2?

emilyrohrbough:
I had the same question [here](https://github.com/cerner/terra-core/pull/329/files#diff-2f4ced636a7ac23a1b1498c37d644ae1R32) on the terra-core PR for 15.5 compatibility. I don't think I would be a bad thing to bump them, but this was the the answer I received from @bjankord was:

> In this specific case, it doesn't matter too much. Using the [caret operator](https://bytearcher.com/articles/semver-explained-why-theres-a-caret-in-my-package-json/#giving-npm-permission-to-install-newer-version) with `^15.5.4` or `^15.4.2` will result npm installing the latest released `15.x` version of react and react-dom.

Matt-Butler:
> Should we bump to react 15.5 instead of 15.4.2?

@mjhenkes, All of our components are currently compatible with react ^15.4.2. We have consumers that are currently using react 15.4.2.


Components should stay with a peerDependency of ^15.4.2 to remain passive.

The root and terra--clinical-site we can set at either 15.4.2 or the latest version. We really want to be able to test with both react 15.4.2 and the latest version of react. We just need to decide which version we default to. What do you think?

mjhenkes:
@Matt-Butler That makes sense to me. We can default to 15.4.2 then.

PR #66, opened 2017-05-22
Adding terra-clinical-modal-manager

tbiethman:
Adding the ModalManager, a Redux-driven modal state manager.


packages/terra-clinical-modal-manager/src/components/ModalManager.jsx
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.updateFullscreenState);
+  }
+
+  updateFullscreenState() {
+    const previousFullscreenState = this.forceFullscreenModal;
+
+    this.forceFullscreenModal = window.innerWidth < getBreakpoints().small;
```
emilyrohrbough: Should this be getBreakpoints().small or should it reflect prop.size?

packages/terra-clinical-modal-manager/src/components/ModalManager.jsx
```diff
@@ -0,0 +1,198 @@
...
+              if (data.preferredType === 'modal' || !app) {
+                openModal(data);
+              } else {
+                app.disclose(data);
+              }
+            },
+          });
+
+          return React.cloneElement(child, { app: childAppDelegate });
+        })}
```
emilyrohrbough: Should this function be broken out for readability? Then this could be reduced to {clonedChildren}

packages/terra-clinical-modal-manager/src/reducers/modalManager.js
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const defaultModalState = Object.assign({}, defaultState, {
+  size: supportedSizes.small,
+});
+
+const modalManager = (state = defaultModalState, action) => {
+  switch (action.type) {
+    case OPEN:
+      return Object.assign({}, open(state, action), { size: action.data.size || supportedSizes.small });
```
emilyrohrbough: Where is `action.data.size` possibly set? I missed it.

emilyrohrbough:
The file name is a bit confusing in the pull request since there are several files named ModuleManger. Could we possibly rename these files to reflect the folder they are contained in?

emilyrohrbough:
Can we add an example where the default size `small` is not used? There are a lot of pieces that need to take size into account and I don't feel like it was tested.

packages/terra-clinical-modal-manager/src/reducers/modalManager.js
```diff
@@ -0,0 +1,43 @@
...
+const supportedSizes = {
+  tiny: 'tiny',
+  small: 'small',
+  medium: 'medium',
+  large: 'large',
+  huge: 'huge',
+};
+
+const defaultModalState = Object.assign({}, defaultState, {
+  size: supportedSizes.small,
```
emilyrohrbough: Should this try for `action.data.size` first?

packages/terra-clinical-modal-manager/src/components/ModalManager.jsx
tbiethman: In past implementations, we've forced the modal to be fullscreen when the window size is below small. We want to do this regardless of the actual desired modal size that the consumer specified. We could theoretically have different versions of the 'fullscreen' that flex based on the desired size, but that's not something we've talked about yet.

packages/terra-clinical-modal-manager/src/components/ModalManager.jsx
tbiethman: It certainly could

packages/terra-clinical-modal-manager/src/reducers/modalManager.js
tbiethman: The `size` value is passed through the disclose API into the `open` action creator for the ModalManager.

For the tests, it originates here: https://github.com/cerner/terra-clinical/blob/terra-clinical-modal-manager/packages/terra-clinical-modal-manager/tests/nightwatch/DemoContainer.jsx#L22

packages/terra-clinical-modal-manager/tests/nightwatch.conf.js
```diff
@@ -0,0 +1,9 @@
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
+
+const testSettings = require('terra-toolkit').testSettings;
+const resolve = require('path').resolve;
+const nightwatchConfiguration = require('../node_modules/terra-toolkit/lib/nightwatch.json');
```
tbiethman: +1

packages/terra-clinical-modal-manager/src/reducers/modalManager.js
tbiethman: This default state will be the initial state of the ModalManager when the application first mounts. There is no action at this point to consider. But it does beg the question whether or not a default is necessary at all...

tbiethman:
@emilyrohrbough (and others) Regarding the file naming question, how would you feel about losing folders altogether and doing something like:
```
terra-clinical-modal-manager/
  src/
    actions.js 
    index.js
    reducers.js
    components/
      ModalManager.jsx
      ModalManagerView.jsx
```

The index.js would tie everything together and be the default export for the package (instead of having the ModalManager container do all that).

emilyrohrbough:
@tbiethman I think the file formatting you listed above is a lot clearer.

packages/terra-clinical-modal-manager/src/ModalManager.jsx
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * Components that will receive the ModalManager's AppDelegate configuration. Components given as children must appropriately handle an `app` prop.
+   **/
+  children: PropTypes.node,
+
+  /**
+   * From `connect`. The Array of component keys that will be used to instantiate the Modal's inner components.
+   **/
+  modalComponentKeys: PropTypes.array,
```
emilyrohrbough: I feel like requiring the consumer to pass in both the keys and array of data with the matching key could make it more complicated/greater room for error within the `buildModalComponents`. Is there a reason not to use just the modalComponentData for building the components?

packages/terra-clinical-modal-manager/src/ModalManager.jsx
```diff
@@ -0,0 +1,198 @@
...
+    this.forceFullscreenModal = window.innerWidth < getBreakpoints().small;
+
+    // Only update the modal if it's minimized, open, and changing states.
+    if (!this.props.isMaximized && this.props.isOpen && previousFullscreenState !== this.forceFullscreenModal) {
+      this.forceUpdate();
+    }
+  }
+
+  buildModalComponents() {
+    if (!this.props.modalComponentKeys || !this.props.modalComponentKeys.length) {
```
emilyrohrbough: Should we check that modalComponentData is not null and is the same length given line 124 below?

packages/terra-clinical-modal-manager/src/ModalManager.jsx
```diff
@@ -0,0 +1,198 @@
...
+      return null;
+    }
+
+    return this.props.modalComponentKeys.map((componentKey, index) => {
+      const componentData = this.props.modalComponentData[componentKey];
```
emilyrohrbough: What happens if modalComponentData does not contain the key? Codes this cause an error in line 126?

packages/terra-clinical-modal-manager/src/ModalManager.jsx
tbiethman: That API matches the state schema directly. Let me see if I can make this better.

packages/terra-clinical-modal-manager/.npmignore
```diff
@@ -0,0 +1,3 @@
+*.log
+node_modules
+src
```
ryanthemanuel: I would add target here as well.  Avoid publishing night watch images

packages/terra-clinical-modal-manager/tests/jest/ModalManager.test.jsx
```diff
@@ -0,0 +1,251 @@
+import React from 'react';
+import AppDelegate from 'terra-clinical-app-delegate';
+import ModalManager from '../../src/ModalManager';
+
+const TestContainer = () => (
+  <div>Hello World</div>
```
ryanthemanuel: I would prefer something more creative here.

packages/terra-clinical-modal-manager/.npmignore
tbiethman: I did do that in a commit below.

packages/terra-clinical-modal-manager/tests/jest/index.test.js
```diff
@@ -0,0 +1,173 @@
...
+      isOpen: 'IS_OPEN_VALUE',
+      isMaximized: 'IS_MAXIMIZED_VALUE',
+      size: 'SIZE_VALUE',
+    };
+
+    expect(mapStateToProps(state)).toEqual(expectedResult);
+  });
+});
+
+// const mapDispatchToProps = dispatch => ({
```
Matt-Butler: nit: remove

tbiethman: D'oh

Issue #77, opened 2017-05-30
Heroku PR deployed issue


windse7en:
## Summary
The Heroku review app can't be deployed due to the code size (after installing all npm modules, around 1.6 GB). It exceeds Heroku [slug size](https://devcenter.heroku.com/articles/limits#slug-size) after the compression.

The root cause is `terra-toolkit`(which is around 130 MB) has been installed repeatedly.

# Enhancement Request
Reorganize the source code and devDependencies to only import `terra-toolkit` once. 

## Expected Behavior
Only import `terra-toolki` once. Reduce the code size less than 1G.

windse7en:
Leana provides a `--hoist` option for bootstrap, which can install common devDependencies only once in [link](https://github.com/lerna/lerna#--hoist-glob). By this way, we will install `terra-toolkit` only once.

But `terra-toolkit` relies on the relative location in the repo. In that case, we need to update `terra-tookit` to remove the relative path related code.

emilyrohrbough:
Resolved in PR #80.

PR #80, opened 2017-06-01
Updates to terra-clinical to fix the Heroku slug size of the project

ryanthemanuel:
### Summary
This PR will fix issue #77  It moves terra-toolkit to be defined as a dependency at the top level only.


packages/terra-clinical-action-header/tests/nightwatch.conf.js
```diff
@@ -1,7 +1,9 @@
 /* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
```
emilyrohrbough: Can we remove this line?

packages/terra-clinical-action-header/package.json
```diff
@@ -21,9 +21,6 @@
     "url": "https://github.com/cerner/terra-clinical/issues"
   },
   "homepage": "https://github.com/cerner/terra-clinical#readme",
-  "devDependencies": {
-    "terra-toolkit": "^0.x"
-  },
```
Matt-Butler: Do we want to do this in terra-core as well? And update the generator?

ryanthemanuel: We should definitely update the generator.  We can't do it in terra-core.  It's a chicken and an egg kind of thing unfortunately.  Can't depend on something at the root level if it's a package

Matt-Butler:
I am curious, we have `nightwatch.conf.js` the same in each package. Can we instead just reference a `nightwatch.conf.js` in the root? I don't want this to hold up the review, maybe we could just log it as an issue.

emilyrohrbough:
@Matt-Butler So I have been wondering about that. I actually think we have some dead code in both core and clinical. From what I can tell, the `nightwatch.conf.js` file is never referenced from a package level, nor from the root. I think it is only being referenced from the site package. It might be something we want to check out.

packages/terra-clinical-action-header/tests/nightwatch.conf.js
ryanthemanuel: Good call.  I'll remove the extraneous ones.

emilyrohrbough:
Slug size is now 113M with this change.

packages/terra-clinical-action-header/tests/nightwatch.conf.js
ryanthemanuel: Done

PR #83, opened 2017-06-02
Integrate terra-i18n-plugin into webpack config.

emilyrohrbough:
### Summary
With `terra-base` v0.5.0 released, the `terra-i18n-plugin` needed incorporated into terra-clinical-site webpack configuration. Without the integration, errors were thrown and the script `npm run start` failed to compile. 

@cerner/terra

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,3 +2,4 @@ lib
 node_modules
 coverage
 build
+**/aggregated-translations/*.*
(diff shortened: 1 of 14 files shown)


emilyrohrbough:
Looks like Travis is failing due to a timeout issue with modal-manager. Timeout occurs after 350 ms for test `Discloses within the modal when modal is already presented` because the element has is not present. Running the tests on my local machine, this test passes with the element taking 771 ms to become present.

mjhenkes:
@emilyrohrbough Do we know what changed to cause the increased wait time?

emilyrohrbough:
@mjhenkes I have no idea.

PR #94, opened 2017-06-14
Remove SCSS theming capability

### LINKED ISSUE #92: Remove SCSS theming capability from clinical components
# Issue Description
We currently support theming via SCSS variables in various components. As we plan to switch to a css-in-js solution for styling, the current implementation of theming variables leaks implementation details of how we are styling/theming components. 

With css-in-js, we'd plan to use JSON files for theming. Switching from SCSS variables to a JSON file for theming would be a non-passive change for our consumers.

We've discussed removing theming functionality via SCSS variables before the v1 release of the various components. This will allow use to roll out theming functionality via JSON files as a new feature rather than a non-passive change.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
All components use hard-coded values instead of SCSS theming variables.

## Current Behavior
Various terra components use SCSS theming variables.

bjankord:
### Summary
Resolves #92 

@cerner/terra

### DIFF
--- a/packages/terra-clinical-detail-view/src/DetailView.scss
+++ b/packages/terra-clinical-detail-view/src/DetailView.scss
@@ -17,13 +17,13 @@
 
 .terraClinical-DetailView-footerText {
   @include terra-text-align-center();
-  color: $terra-grey-50;
+  color: #909496;
   font-size: 1rem;
   line-height: 1.428;
 }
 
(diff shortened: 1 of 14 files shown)


mjhenkes:
@bjankord thanks for the PR

PR #123, opened 2017-07-06
WIP Danger integration

bjankord:
### Summary
esting [danger](http://danger.systems/js/) integration. This uses the terra-bot and danger to remind people to update the changelog whenever they modify files in the `src` directory within a package. It will also warn people about big PRs (more than 100 additions/deletions) and suggest breaking them into smaller PRs.

### Original files that were modified
screen shot 2017-07-06 at 11 47 35 am

### Comment from bot noting PR needs a changelog entry. This is based on src files being modified.
This also blocks CI from passing until a changelog entry is added when modifying src files.
screen shot 2017-07-06 at 10 14 11 am

## And then the build passes
screen shot 2017-07-06 at 11 45 56 am

@cerner/terra


dangerfile.js
```diff
@@ -0,0 +1,26 @@
...
+const modifiedSrcFiles = danger.git.modified_files.filter((filePath) => {
+  const srcFilePattern = /^packages\/([a-z-])*\/src/i;
+  return srcFilePattern.test(filePath);
+});
+
+const hasCHANGELOGChanges = modifiedChangelog.length > 0;
+const hasModifiedSrcFiles = modifiedSrcFiles.length > 0;
+
+// Fail if there are src code changes without a CHANGELOG
+if (hasModifiedSrcFiles && !hasCHANGELOGChanges) {
```
emilyrohrbough: Out of curiosity, does danger.git.modified_files return the number of file changes? I wonder if there would ever be a case where we would have both `hasModifiedSrcFiles` & `hasCHANGLELOGChanges` but it didn't capture the full scope of the changes.

bjankord: `danger.git.modified_files` returns an array of strings of the file path + file name of files that have been modified.

The below code is used to filter that array down to only modified files with in the `src` directory of a package. 

The thinking behind this is that there will be trivial changes to docs/tests/etc that likely don't need changelog entries. For those, I'd recommend people use their best judgement. For changes to files in the src directory though, we should be documenting those.

```
const modifiedSrcFiles = danger.git.modified_files.filter((filePath) => {
  const srcFilePattern = /^packages\/([a-z-])*\/src/i;
  return srcFilePattern.test(filePath);
}
```

PR #145, opened 2017-07-12
Migrate Item View to CSS Modules.

emilyrohrbough:
### Summary
Addresses issue #133

### DIFF
--- a/packages/terra-clinical-item-view/CHANGELOG.md
+++ b/packages/terra-clinical-item-view/CHANGELOG.md
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Converted component to use CSS modules
 
 1.0.0 - (June 28, 2017)
 -----------------
--- a/packages/terra-clinical-item-view/package.json
+++ b/packages/terra-clinical-item-view/package.json
@@ -26,16 +26,14 @@
     "react-dom": "^15.4.2",
     "terra-base": "^1.0.0",
     "terra-clinical-item-display": "^1.0.0",
-    "terra-icon": "^1.0.0",
-    "terra-mixins": "^1.0.0"
+    "terra-icon": "^1.0.0"
   },
   "dependencies": {
     "classnames": "^2.2.5",
     "prop-types": "^15.5.8",
     "terra-base": "^1.0.0",
     "terra-clinical-item-display": "^1.0.0",
-    "terra-icon": "^1.0.0",
-    "terra-mixins": "^1.0.0"
+    "terra-icon": "^1.0.0"
   },
   "scripts": {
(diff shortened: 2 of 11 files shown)


PaulSolDev:
item view ie 11

At small (phone sized) width text is being cut-off with no indication of truncation.  Noticed on all browsers.

PaulSolDev:
item view accessory

Content is cut-off on IE and Safari.

emilyrohrbough:
> At small (phone sized) width text is being cut-off with no indication of truncation. Noticed on all browsers.

The ItemView requires consumers to provide the item displays and the item display styles desired. 

Item Displays do not apply truncation styling on default, so no truncation would not be expected here. To get truncation on these examples,  'isTruncated` would need to be added to each display [here](https://github.com/cerner/terra-clinical/pull/145/files#diff-15693ed03572e628e3d2833a94db61f3R4).

emilyrohrbough:
> Content is cut-off on IE and Safari.

I checked out the issue you pointed out. The element in concern there is a max-height of 60 px and a max-width of 60 px applied to the accessory. The image alt-text "Graphic Start" and "Graphic End" were used in place of an actual image in these examples. These appear to be difference in how browser alt-text is rendered, but the styling applied do appear to be the same on the elements when inspecting the browser.

PaulSolDev:
I may be misunderstanding the element, but from a functional perspective we can't cut-off content without indicating that information is not being displayed (common CHIA).  Our two methods have been to wrap extra content to fit, or truncate and show the ellipsis.  Can I assume that the correct behavior for this element when truncation is not applied should be to wrap the text while fully displaying the content?

PaulSolDev:
Would the tests around the accessory possible by improved by including a graphic option?  Either way, should we be concerned that different browsers have the potential to display the same content differently.  This is similar to my last comment in that we always worry about losing information without notifying a user.

emilyrohrbough:
@PaulSolDev 

> from a functional perspective we can't cut-off content without indicating that information is not being displayed (common CHIA). Our two methods have been to wrap extra content to fit, or truncate and show the ellipsis

I looked further into this. It does look like the example you posted is applying our truncation, when this tag is removed it will wrap the text.  I agree, cutting the text off is incorrect.

> Would the tests around the accessory possible by improved by including a graphic option?

Yes adding a graphic would be good for testing consistency among browser comparison and for adding context to the tests.

>  should we be concerned that different browsers have the potential to display the same content differently.

The content that is being cutoff is the fallback display if the image does not load. This would be something we should ask UX about.

I will log an issue addressing each thing: 1) truncation show an ellipsis 2) Adding graphic to action header tests and 3) Addressing fallback text styling inconsistencies.

PaulSolDev:
Functional review complete.

PR #163, opened 2017-07-26
Fix the Display Content Cell Creation

emilyrohrbough:
### Summary
Michael McCurdy reached out to ask for an example that showed alignment was maintained in table view when an item did not have a comment present. During this update, I discovered the itm display table cell creation was not locked into the expected number of item display columns determined by the first item row. 

This PR fixes the table cell creation to item displays and adds and example to prove alignment.

@cerner/terra


PaulSolDev:

Issue #142, opened 2017-07-11
Uplift Header to include a Subheader option


claytonbeese:
# Issue Description
Adding a flag onto the Header control to allow it to be a 'subheader'.  Flag would change the background to grey-25, and possibly shrink the vertical padding of the header to differentiate it.  Needed for the Radiology worklist.

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Should behave exactly as a header.  Visual distinction will be used only to indicate a secondary level of hierarchy.

## Current Behavior
Header is a static background color and padding height, leading to a single level of hierarchy.

claytonbeese:
![image](https://user-images.githubusercontent.com/28354462/28188076-774c2896-67e6-11e7-8255-044fd06ed801.png)

attaching an initial mockup of the subheader

nagrawal3:
# Tech Design for terra-clinical-header
## Summary:
Terra-clinical-header will accept another flag isSubsection to make the header as sub-sectional header. To distinguish sub-sectional header with the header, background color of the sub-sectional header will be @grey25 i.e. #c8cacb and it will not have any border.

To keep this as passive change, isSubsection prop will have a default value as false.

## React Props:
| Prop     | Type | Default | Description  |
|----------|------|---------|--------------|
| startContent | element | null    | Content to be displayed at the start of the header, placed before the title |
| title | string | ''    | Text to be displayed as the title in the header bar |
| endContent | element | null    | Content to be displayed at the end of the header |
| isSubsection | Bool | false    | Should header be a sub-sectional header |

## Example:
```
import React from 'react';
import Button from 'terra-button';
import Header from 'terra-clinical-header';

const SubSectionalHeaderExampleComponent = () => (
  <div>
    <Header
      startContent={<div><Button text="Button" style={{ margin: '0 10px 0 0' }} /></div>}
      title="Default Header"
      endContent={<div><Button text="Button" style={{ margin: '0 0 0 10px' }} /></div>}
      isSubsection
    />
  </div>
);

export default SubSectionalHeaderExampleComponent;
```

emilyrohrbough:
Nit: can we use the name `isSubheader` instead of `isSubction`

nagrawal3:
I can use `isSubheader` as the prop.

AS per UX design, font size will be same as we have for header.

tbiethman:
[...]

PR #164, opened 2017-07-27
Uplift Header to include a Subheader option

nagrawal3:
### Summary
Resolves: [#142](https://github.com/cerner/terra-clinical/issues/142)
This will add an option to render Subheader through Header component

@cerner/terra


packages/terra-clinical-header/src/Header.jsx
```diff
@@ -22,17 +22,24 @@ const propTypes = {
    * Content to be displayed at the end of the header
    */
   endContent: PropTypes.element,
+
+  /**
+   * A Boolean indicating if element is a subheader. Default is false
```
JakeLaCombe: Nitpick.

Do we need to specify default is false in the prop documentation? I believe our table will latch onto the defaultProps and get the value from there.

packages/terra-clinical-site/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Unreleased
 
 ### Changed
 * Update Header example
+* Added Subheader example
```
emilyrohrbough: Can we update the changelog to read something like
`### Added`
`* isSubheader boolean prop to allow rendering of subheader component`

packages/terra-clinical-site/src/examples/header/Subheader.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import Button from 'terra-button';
+import Header from 'terra-clinical-header';
+
+const Subheader = () => (
+  <div>
+    <Header
+      startContent={<div><Button text="Button" style={{ margin: '0 10px 0 0' }} /></div>}
+      title="Default Header"
```
emilyrohrbough: nit: default subheader

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -47,8 +54,14 @@ const Header = ({ title, startContent, endContent, ...customProps }) => {
     headerContent = titleElement;
   }
 
+  if (isSubheader) {
```
Matt-Butler: You can clean this up if you want to:
```
const headerClass = isSubheader ? 'subheader' : 'header';
```

packages/terra-clinical-site/CHANGELOG.md
nagrawal3: @emilyrohrbough just to be sure before I push my changes:

should this file say something like this:

```
### Added
* Example to render subheader component
```
as we are only adding example file in terra-clinical-site.
and CHANGELOG.md for terra-clinical-header should say something like below:

```
### Added
* isSubheader boolean prop to allow rendering of subheader component
```

emilyrohrbough: Yes good catch.

PaulSolDev:
In IE 10 and 11 the subheader does not resize correctly at small/extra small sizes.  Content on both left and right disappears and the text that is visible fails to truncate.  This same issue did not occur in Microsoft Edge.

ie 11 subheader

leemurdock:
Everything looks good in Safari.

neilpfeiffer:
UX needs to re-review the color/contrast of background color and buttons in the subheader, will post an update after having a discussion.

dkasper-was-taken:
Logged issue #173 , it's not related to this PR for subheader styling.

tbiethman:
@neilpfeiffer Any update on the contrast/color review?

neilpfeiffer:
Thanks @tbiethman and @dkasper-was-taken. After reviewing, I am recommending to UX not to use buttons in the subheader as startConent and endContent for this current theme/styling (We will revisit when someone on UX requests that they need it). 

Let's update the "Subheader with Content" example to use the terra-icon "briefcase" for the startContent, and terra-clinical-item-display secondary style with the text "May 01, 2011" as the endContent.

-----

Tested: PR #164 

**Reviewed:**
- Validated Subheader styling enhancement added to terra-clinical-header.



**Pass** |
-----------|
UX Review|

tbiethman:
@neilpfeiffer Are you recommending _any_ type of content to be used in start/endContent for the Subheader? If you aren't, the Subheader really needs to be a separate component that doesn't expose those props.

neilpfeiffer:
@tbiethman As the API currently stands, start/endContent accept any type of content, so we'll keep it the same for subheader. We'll keep it understood for UX/best practices that you don't drop in certain content (alerts, no-data glyphs, etc.) and for the short-term include buttons on that list based on color, even though it technically is allowed. 

Let's make sure @PaulSolDev has a chance to do a final functional reviewed (& remove the label "needs functional review") and make sure we get the requirements updated so we can remove the "needs design input" label from the issue #142 before merging.

PaulSolDev:
@neilpfeiffer What requirements do you believe need to be updated?

neilpfeiffer:
@PaulSolDev I'm not sure what Tim has been doing on your team's side for creating or updating requirements (might be good to have you give me a quick walkthrough) or to what detail. Since this adds a new subheader "type" that wasn't there before, didn't know if you also needed to capture it as well in functional requirements. (My previous comment was just to allow all the dotting i's, crossing t's process wasn't getting skipped)

PaulSolDev:
@neilpfeiffer We are housing our requirements for like components (view models/react components) in the same module and calling out the various differences.  In looking at Tim's previous requirements around the Header element there is already a functional requirement around the option to set at subheading.  Even though this was initially written for the View Model I believe we should still be covered as we typically point out variance between the two technologies when they exist, and I'm not seeing any such documentation.  Tim is back on Monday so I will confirm this with him then.

mjhenkes:
@nagrawal3 could you take a look at fixing the conflicts in the branch?

leemurdock:
Everything looks good in all browsers.

PR #167, opened 2017-07-31
Terra clinical embedded content consumer

### LINKED ISSUE #70: Create EmbeddedContentConsumer component to interface with embedded iframe content
## Progress
- [ ] Design Input
- [ ] Expected Behaviour
- [ ] Tech Design
- [ ] Development
- [ ] Functional Validation
- [ ] UX Validation
- [ ] Released

------------------------------------------------------------

# Feature / Enhancement Request

Future React applications will need to embed iFrame content into themselves in order to present non-React content. However, we still want content in those embedded frames to communicate to their containing applications (using a messaging system like XFC) in order to drive workflows.

I would like to develop a generic EmbeddedContentConsumer component that will mount an iFrame within itself using XFC and handle the generated messages in a standard way.

## Expected Behavior

I have an (admittedly out of data) example in this branch: https://github.com/cerner/terra-clinical/blob/pdf_redux/packages/terra-clinical-application/src/patient-app/embedded-content-consumer/EmbeddedContentConsumer.jsx

Ideally, the EmbeddedContentConsumer (better names welcome :D) would have a simple prop API:
* `src` - The URL for the iFrame's `src` attribute
* `handlers` - An Object mapping message names to function handlers

When the EmbeddedContentConsumer mounts, it will mount the iFrame using XFC and establish communication with the Provider in the iFrame content. Then, when the Provider dispatches messages back to the EmbeddedContentConsumer, the EmbeddedContentConsumer will use the message name to lookup the proper handler from the `handlers` prop and will call the handler with the message payload as a parameter.

This would allow more opinionated components to wrap the EmbeddedContentConsumer and provide specific URLs and handlers. We may wish to include some default handlers in the EmbeddedContentConsumer itself, but I'm not sure what those defaults would be at this time.

The above linked branch also includes a EmbeddedContentProvider: https://github.com/cerner/terra-clinical/blob/pdf_redux/packages/terra-clinical-application/src/patient-app/embedded-content-provider/EmbeddedContentProvider.jsx

Theoretically, React content in the iFrame could be wrapped in that component and have the bootstrapping handled automatically. Non-React content would have to handle the bootstrapping separately (but in a similar way). I'm not sure a component like EmbeddedContentProvider is necessary at this time, but I'm documenting the prototype here to hopefully give everything a bit more context.

rasthana:
### Summary
An Embedded content consumer component to interface with embedded iframe content. This pull request includes the initial component API. Fixes #70


packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,51 @@
...
+  src: undefined,
+  eventHandlers: {},
+};
+
+class EmbeddedContentConsumer extends React.Component {
+
+  componentDidMount() {
+    if (this.embeddedContentWrapper) {
+      Consumer.init();
+      this.xfcFrame = Consumer.mount(this.embeddedContentWrapper, this.props.src);
```
tbiethman: Depending on how fancy we want to get out of the gate, we might want to expose all the Consumer mount options: https://github.com/cerner/xfc#setting-up-a-consumer

Basically it'd be adding secret and resizeConfig support.

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.scss
```diff
@@ -0,0 +1,17 @@
+@import '~terra-mixins';
```
emilyrohrbough: With the conversion to CSS Modules, we have included a webpack plugin that handles bi-directional styling. `@import '~terra-mixins';` can be removed here and the terra-mixing package dependencies can also be removed.

packages/terra-clinical-embedded-content-consumer/docs/README.md
```diff
@@ -0,0 +1,28 @@
...
+  source="content url"
+  eventHandlers={ 'handlerKey': () => {} };
+/>
+```
+
+## Component Features
+* [Cross-Browser Support](https://github.com/cerner/terra-core/wiki/Component-Features#cross-browser-support)
+* [Responsive Support](https://github.com/cerner/terra-core/wiki/Component-Features#responsive-support)
+* [Mobile Support](https://github.com/cerner/terra-core/wiki/Component-Features#mobile-support)
+* [Internationalization Support](https://github.com/cerner/terra-core/wiki/Component-Features#internationalization-i18n-support)
```
emilyrohrbough: The Internationalization and Localization support can be removed since this component doesn't need translations.

packages/terra-clinical-embedded-content-consumer/tests/nightwatch.conf.js
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
emilyrohrbough: This file can be removed with the newest toolkit version.

packages/terra-clinical-embedded-content-consumer/tests/jest/EmbeddedContentConsumer.test.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import EmbeddedContentConsumer from '../../src/EmbeddedContentConsumer';
```
emilyrohrbough: oops want to import from lib

emilyrohrbough:
Do we want to worry about updating this to css modules now?

mjhenkes:
@emilyrohrbough Yes we should update to css modules for this PR

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,51 @@
...
+const propTypes = {
+  /**
+   * The source URL of the content to load.
+   */
+  src: PropTypes.string.isRequired,
+  /**
+   * An object containing event handlers keyed by the event name.
+   * (e.g. {'eventA': function() {}, 'eventB': function() {}})
+   */
+  eventHandlers: PropTypes.object,
```
tbiethman: Can you also define the shape of the eventHandlers object using proptypes?

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+const defaultProps = {
+  src: undefined,
+  eventHandlers: {},
+};
+
+class EmbeddedContentConsumer extends React.Component {
+
+  componentDidMount() {
+    if (this.embeddedContentWrapper) {
```
tbiethman: I think I had this in the prototype, but the if check here isn't necessary. If the div didn't mount before componentDidMount is called, we'd be having bigger problems.

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.scss
```diff
@@ -0,0 +1,15 @@
+.terraClinical-EmbeddedContentConsumer {
+  height: 100%;
+  width: 100%;
+
+  .xfc { // stylelint-disable-line selector-class-pattern
```
tbiethman: When you move to css modules, you're going to have to mark this '.xfc' selector as global to prevent css modules from converting it.

{code}
:local {
  .container {
    height: 100%;

    :global {
      .xfc {
        height: 100%;
     
        iframe {
          border: 0;
          height: 100%;
          width: 100%;
        }
      }
    }
  }
}
{code}

Something like that should work.

packages/terra-clinical-embedded-content-consumer/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "terra-clinical-embedded-content-consumer",
+  "main": "lib/EmbeddedContentConsumer.js",
+  "version": "0.0.0",
+  "description": "The Embedded Content Consumer is the applicaiton component which is embedding content within it.",
```
mschile: Misspelled 'application'.  Suggest providing an example.

packages/terra-clinical-embedded-content-consumer/package.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "homepage": "https://github.com/cerner/terra-clinical#readme",
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-base": "^0.x",
```
mschile: Update to ^1.0.0

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,51 @@
...
+  src: PropTypes.string.isRequired,
+  /**
+   * An object containing event handlers keyed by the event name.
+   * (e.g. {'eventA': function() {}, 'eventB': function() {}})
+   */
+  eventHandlers: PropTypes.object,
+};
+
+const defaultProps = {
+  src: undefined,
```
mschile: Won't src automatically be defaulted to undefined?

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: Planning on changing to the following, thoughts?
```
eventHandlers: PropTypes.arrayOf(PropTypes.shape({
  key: PropTypes.string,
  handler: PropTypes.func,
})),
```

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+class EmbeddedContentConsumer extends React.Component {
+
+  componentDidMount() {
+    Consumer.init();
+    this.xfcFrame = Consumer.mount(this.embeddedContentWrapper, this.props.src);
+
+    if (this.props.eventHandlers) {
+      this.props.eventHandlers.forEach((event) => {
+        this.xfcFrame.on(event.key, event.handler);
```
mhemesath: do you want to bind the event handler to the xfcFrame?

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+class EmbeddedContentConsumer extends React.Component {
+
+  componentDidMount() {
+    Consumer.init();
+    this.xfcFrame = Consumer.mount(this.embeddedContentWrapper, this.props.src);
+
+    if (this.props.eventHandlers) {
+      this.props.eventHandlers.forEach((event) => {
```
mhemesath: So, I suggest that this component has first class support for all 1sst class XFC events. Then I'd have this take a generic customEvent handler. The consumer of this component can then handle custom Events as it wishes.

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
tbiethman: Are you asking whether we want to bind the specific 'event.handler' function directly to the frame vs. wrapping it in another function or something? Or is there another place to listen to events coming from the Provider?

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
tbiethman: So something like this?
```jsx
<EmbeddedContentConsumer
  src="www.example.com"
  onMount={() => { console.log('mounted'); }}
  onLaunch={() => { console.log('launched'); }}
  onAuthorize={() => { console.log('authorized'); }}
  onFullscreen={() => { console.log('provider wants to go fullscreen'); }}
  customEvents={{
    customEvent1: () => { console.log('customEven1 was triggered'); }
    customEvent2: () => { console.log('customEven2 was triggered'); }
  }}
```
I'm not sure how we can get away with having a single customEvent handler (unless we could do a wildcard-like bind on the frame to catch everything that's not an xfc-specific event).

jmsv6d: The customEvents prop could have the same structure as the `eventHandlers` prop has now.

tbiethman: @jmsv6d Whoops you're right, I'll update the example.

packages/terra-clinical-embedded-content-consumer/README.md
```diff
@@ -0,0 +1,32 @@
+# Terra Clinical Embedded Content Consumer Site
```
emilyrohrbough: nit: remove site

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
mhemesath: Also, XFC uses event emitter so this lets you do the following
* Use regex to catch events 
* Pass a single object and bind all events at once

https://github.com/Olical/EventEmitter/blob/master/docs/api.md#addlisteners

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,45 @@
...
+  eventHandlers: PropTypes.arrayOf(PropTypes.shape({
+    key: PropTypes.string,
+    handler: PropTypes.func,
+  })),
+};
+
+class EmbeddedContentConsumer extends React.Component {
+
+  componentDidMount() {
+    Consumer.init();
```
kafkahw: Consumer.init() should be called only once in your application, so you might want to put it in a higher level component. You probably don't want to call it for every embedded content.

packages/terra-clinical-embedded-content-consumer/src/EmbeddedContentConsumer.scss
```diff
@@ -0,0 +1,20 @@
+/* stylelint-disable max-nesting-depth */
+/* stylelint-disable selector-class-pattern */
+/* stylelint-disable selector-max-compound-selectors */
+:local {
+  .container {
```
mhemesath: @mjhenkes mentioned our team may want to use this component. If we do, these styles will have to be optional or moved to a wrapper component. Our embedded components are not 100% height.

rasthana:
Closing this PR since it will be moved to terra-core: https://github.com/cerner/terra-core/pull/737

PR #170, opened 2017-08-01
Migrate Application Site to CSS Modules

### LINKED ISSUE #127: Convert Application Site to CSS Modules
# Issue Description
To better align with modern CSS practices and modern CSS theming capabilities, we want to update this component to use CSS modules and if applicable, CSS custom properties.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Component styles are converted to use CSS modules. If applicable, convert SCSS variables to CSS custom properties.

## Current Behavior
Component styles use manual name spaced classes and SCSS variables.

emilyrohrbough:
### Summary

-Migrates Application-Site to CSS Modules. Closes #127.
-Fixes application-site demo styling in terra-clinical-site.

@cerner/terra


emilyrohrbough:
@bjankord @mjhenkes Build is failing due to ItemCollection nightwatch tests. With the conversion to Css Modules, we discussed replacing any class that did not apply styling as a `data-class`. In creating the table view, ItemCollection tried to apply data-classes, which is applied as customAttr. However, the TableCell component already applies a data-class, which is override the ItemCollection, which is causing tests to fail.

packages/terra-clinical-site/src/site.scss
```diff
@@ -25,7 +25,8 @@
       height: 100%;
     }
 
-    [class*="ItemView__accessory"] {
+    [class*="ItemView__accessory"],
+    [class*="ItemView__accessory"] > * {
```
StephenEsser: This seems a little out of place, what is this doing exactly?

This wildcard class selector is dependent on the css-loader options defined in the webpack config. If the option changes this css would stop working.

packages/terra-clinical-application-site/src/index.scss
```diff
@@ -16,6 +16,7 @@ body {
   height: 100%;
 }
 
-.terraClinical-ItemView-accessory {
+[class*="ItemView__accessory"],
+[class*="ItemView__accessory"] > * {
```
StephenEsser: This seems a little out of place, what is this doing exactly?

This wildcard class selector is dependent on the css-loader options defined in the webpack config. If the option changes this css would stop working.

emilyrohrbough: The Item View styling gives accessory items a max-width of 60px. The accessory provided in the demo exceeds this maximum width, which is why the styling is being overridden.

StephenEsser: This is really brittle css, but I understand it's just a conversion. It might be worth investigating if the max-width constraint of the accessory should be changed in the ItemView itself. It feels awkward that we can't use it without custom CSS.

emilyrohrbough: I agree, it seems like a tight constraint and it is brittle css. This is something Tim & I discussed at on point on the requirements side when I was working on item-collection.

PaulSolDev:
Spoke with @tbiethman and @emilyrohrbough and it was decided this does not need functional review.  Label removed accordingly.

PR #174, opened 2017-08-03
Header text truncation

### LINKED ISSUE #173: Header text is not truncating properly on IE 10 and IE 11
# Issue Description
Header text is not truncating properly on IE 10 and IE 11. This is also linked to the flex issues in the underlying terra-arrange implementation.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

dkasper-was-taken:
### Summary
Remove arrange from header, and add examples of truncated text.

Fixes #173


emilyrohrbough:
Do we want to put these examples/test cases in the nightwatch test examples?

emilyrohrbough:
Why did we remove header from arrange? Just curious.

dkasper-was-taken:
Headers implementation is too rigid to allow for the additional css styles needed to support text truncation across all browsers. Additional styling needs to be placed on it's sub-components, but that styling would likely break the existing implementation, so it'd be a subset of configuration, and I believe that there was a good deal of push break to keep arrange very basic.

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -31,25 +30,36 @@ const defaultProps = {
...
+
+  let endElement;
+  if (endContent) {
+    endElement = <div className={cx('flexEnd')}>{endContent}</div>;
   }
 
   return (
     <header {...customProps} className={cx('header', customProps.className)}>
-      {headerContent}
+      <div className={cx('flexContainer', customProps.className)}>
```
StephenEsser: It looks like customProps.className is getting spread onto both the outer and inner flex container. Would it simplify anything if the css from `flexContainer` was applied to `header` directly? Then the extra div could be removed.

dkasper-was-taken: I think that was a copy/paste mistake. I'll update it.

packages/terra-clinical-site/src/examples/header/HeaderLongText.jsx
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable no-alert */
+import React from 'react';
+import Header from 'terra-clinical-header';
+
+const LongTextWithButtons = () => (
+  <div>
+    <br />
+    <Header
+      title="Header Header Header Header Header Header Header Header Header Header Header Header Header Header Header Header Header Header"
```
StephenEsser: For these long text examples would you be against using Lorem Ipsum text?

http://generator.lorem-ipsum.info/_latin

dkasper-was-taken:
@emilyrohrbough I updated a site example for long text with content section.

emilyrohrbough:
@dkasper-was-taken The nightwatch test looks good. With this suggestion I mean it might be good to add these to the nightwatch examples and remove them from the site page examples. It was discussed at one point that we want to use site to display the expected use of the component while the nightwatch examples would be used to help demonstrate edge case examples.

PaulSolDev:
It appears that header text can be cut off without any indication of truncation at small sizes.  I believe our existing patterns state we must show when content is truncated which would lead me to expect the left and right content to adjust accordingly (should leave at least one character and the ellipsis visible at all times).  

ie 10 header truncation

mikertr:
I'd suggest completely omitting the header title if you can't at least get a full character plus an ellipsis.  Also, it looks like there are different truncation rules for "header with content" and "long text with buttons".  Is that purposeful?  
screen shot 2017-08-09 at 2 25 56 pm

dkasper-was-taken:
@mikertr What browser are you using that you see that discrepancy?

mikertr:
@dkasper-was-taken here's an example from macOS Firefox (header with left and right content).  As the space for header title shrinks, the truncation works until you cross under the threshold of 1 character + ellipsis.  Once it gets smaller than that, there's no point in displaying header text at all.
screen shot 2017-08-09 at 4 39 18 pm
screen shot 2017-08-09 at 4 39 07 pm

dkasper-was-taken:
@mikertr After some discussion, I believe this behavior will just be documented, we cannot easily determine a minimum size, as theming w/font sizes make accounting for the size of the characters inconsistent.

leemurdock:
Looks good in Safari.

PaulSolDev:
Functional review complete. No issues identified.

PR #180, opened 2017-08-15
Update site to use slide panel

emilyrohrbough:
### Summary
Update the display of terra-site to use a slide-panel. Sync with https://github.com/cerner/terra-core/pull/720.

@cerner/terra


dkasper-was-taken:
Your my hero. Thanks for doing this one.

PR #188, opened 2017-08-31
Update components to use terra-base 2.0.0

bjankord:
### Summary
Updating all components that had a dependency on base to use terra-base@2.0.0.
Added margin resets to elements used in clinical components which used to have margin reset in terra-base@1.0.0

@cerner/terra

### DIFF
--- a/packages/terra-clinical-action-header/CHANGELOG.md
+++ b/packages/terra-clinical-action-header/CHANGELOG.md
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Change dependency on terra-base to v2.0.0
 
 1.2.0 - (August 16, 2017)
 -----------------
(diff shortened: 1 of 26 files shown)


packages/terra-clinical-action-header/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Change dependency on terra-base to v2.0.0
```
StephenEsser: Maybe not a huge deal, but I believe this update caused all the tests links to receive bullets

jmsv6d: That's because we're using `<li>` tags and not the List component

packages/terra-clinical-detail-view/src/DetailView.scss
```diff
@@ -15,7 +15,9 @@
 
   .primary-text {
     font-size: 1.428rem;
+    margin-bottom: 0;
     margin-top: 0;
+
```
emilyrohrbough: nit: extra line

Issue #169, opened 2017-08-01
Visual Inconsistencies in Item Collection 


emilyrohrbough:
# Issue Description
The wrapper component item collection provided the user with two views: table and list view. The display generation between the two views are different, so when varying elements are supplied, it can result in inconsistent visuals.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
From a functional perspective, what information is displayed and how it appears on the screen should be consistent. 

## Current Behavior
The table generation handles this by using the first row supplied and the following rows will display the same number of columns. If more items are provided than the first, it will not show the additional items, and if less is provided, an empty cell is shown. It results in the the following display: 

![screen shot 2017-08-01 at 10 45 35 am](https://user-images.githubusercontent.com/14099737/28834228-4cbe5ba6-76a7-11e7-931a-497dce915885.png)
 
However, when shrunk to a list view, all items are shown and can result in more information than was provided in the table view. Additionally, if some have start accessory and others do not, the content does not visually line up. It results in the following display:

![screen shot 2017-08-01 at 10 45 47 am](https://user-images.githubusercontent.com/14099737/28834334-98f7cc64-76a7-11e7-9d09-958cb883f6bb.png)

## Steps to Reproduce

1. Go to http://engineering.cerner.com/terra-clinical/#/tests/item-collection-tests/varying-items
2. Shirk the page smaller to view the difference between table and list view.

emilyrohrbough:
Closed in #190

PR #190, opened 2017-09-05
Fix Visual Inconsistencies in Item Collection

emilyrohrbough:
### Summary
This PR addresses https://github.com/cerner/terra-clinical/issues/169. 

Additionally, it updates then nightwatch tests to use expect syntax instead of assert.


packages/terra-clinical-item-collection/src/_CreateListView.jsx
```diff
@@ -1,13 +1,30 @@
...
 import ItemView from 'terra-clinical-item-view';
+import styles from './ItemCollection.scss';
 
 function createListView(rows, listStyles, selectedIndex, handleSelection) {
+  const numberOfDisplays = rows[0].displays.length;
+
   const listContent = rows.map((row, rowIndex) => {
+    const { startAccessory, displays, comment, endAccessory, itemStyles } = row;
+
+    const startAccessoryContent = <div className={styles['list-accessory']}>{startAccessory}</div>;
```
jmsv6d: So this always adds a div with that 40px width. What if none of the items in the list have an accessory? Should they all still have the empty 40px in front?

emilyrohrbough: hmm good point. I made some changes in my last commit that simplified the code that took that checkout. (I knew there was a reason I had it initially)  I'll revert those changes.

jmsv6d:
I think this looks good to fix the inconsistencies. I just think that it's kinda weird that we're just using the first row to determine if accessories should be displayed. But maybe a separate issue needs to be logged for that.

emilyrohrbough:
@jmsv6d The use of the first row to determine the layout was something @mjhenkes and I discussed during the design of this component to determine the number of table columns dynamically. I agree its a bit weird though

emilyrohrbough:
@PaulSolDev @tm4027 This is ready for functional verification.

PaulSolDev:
Appears correct in IE 10, 11, and Edge.

leemurdock:
Checks out in Safari.

emilyrohrbough:
@PaulSolDev @tm4027 @mikertr @leemurdock  what validations are left on this?

emilyrohrbough:
@PaulSolDev @mikertr I restarted the heroku deployment. Here is the link: https://terra-clinical-deployed-pr-190.herokuapp.com/static/#/site

mikertr:
Looks good in macOS Chrome and Firefox.

PR #195, opened 2017-09-20
Remove Defined "selector-class-pattern" Rule

### LINKED ISSUE #193: Remove "selector-class-pattern" Rule
# Issue Description
Remove "selector-class-pattern" rule from .stylelintrc file. This rule is no longer needed with the conversion to css-modules complete. With removing this rule, all scss files will need cleaned up to remove `stylelint-disable selector-class-pattern` comments.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

emilyrohrbough:
### Summary
Removes our defined "selector-class-pattern" rule. Closes #193.

In this PR:
- Updated Header's classnames to follow default selector-class-pattern
- Updated Item View's non-prop defined classes. 
- Disabled selector-class-pattern for prop defined classes in ItemDisplay and ItemView. 

**By prop-defined classes:** classNames defined via the prop value. For example: ItemView's layout props accepts `oneColumn` or `twoColumns` which is directly assigned as the class name.


packages/terra-clinical-item-view/src/ItemView.scss
```diff
@@ -27,13 +27,16 @@
     max-width: 60px;
   }
 
+  // Disable selector-class-pattern. Classes defined via assessoryAlignment prop.
+  /* stylelint-disable selector-class-pattern  */
```
jmsv6d: Couldn't we just update how the class is set in the component so we can follow the proper naming convention? So switch `{ [accessory-${accessoryAlignment}]: accesoryAlignment }` to 
`
{
    accessory-align-top: accessoryAlignment === 'alignTop',
    accessory-align-center: accessoryAlignment === 'alignCenter'
}`

emilyrohrbough: Sure thing.

PR #201, opened 2017-10-09
Set webpack dev server at 2.7.1 for IE10 compatibility

mjhenkes:
### Summary
Webpack > 2.7.1 dropped IE 10 & 11 support.
See https://github.com/cerner/terra-core/pull/901
@cerner/terra

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,6 @@
     "terra-props-table": "^1.0.0",
     "terra-toolkit": "^2.1.0",
     "webpack": "^3.6.0",
-    "webpack-dev-server": "^2.9.1"
+    "webpack-dev-server": "^2.7.1"
   }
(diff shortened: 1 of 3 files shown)


mhemesath:
11 still works as it supports ‘let’ and ‘const’

package.json
```diff
@@ -96,6 +96,6 @@
     "terra-props-table": "^1.0.0",
     "terra-toolkit": "^2.1.0",
     "webpack": "^3.6.0",
-    "webpack-dev-server": "^2.9.1"
+    "webpack-dev-server": "^2.7.1"
```
mhemesath: Remove the ^

mjhenkes: derp, good call.

Issue #192, opened 2017-09-14
Make header compatible with the collapsible menu view


jmsv6d:
# Issue Description
The collapsible menu view is going need to be used in headers, the collapsible menu view will take up all space available to it so the header will need the content sections to have a constrained width to keep the collapsible menu view from taking up the entire header. 

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The header should have a max width set on the content sections.

## Current Behavior
The content sections of the header can take up as much space as they want causing the content to expand the entire width of the header.

cwalten:
We are experiencing an issue related to this.

## Expected Behavior
When a collapsible menu is added as a child to an action header, and onBack and onClose methods are provided, all buttons should be visible in the header.

## Current Behavior
The collapsible menu pushes the Close button off of the screen, forcing it to wrap. (see attached screenshot)

<img width="560" alt="screen shot 2017-11-07 at 4 16 58 pm" src="https://user-images.githubusercontent.com/19825717/32521009-938b2802-c3d7-11e7-8727-b3a4ccb87a66.png">

kschuste:
@cwalten I believe this occurs as the div (action_header) containing your contents needs to have it's width be 100% and have a display of 'flex'.  This should allow the button to layout next to each other.
<img width="1007" alt="screen shot 2017-11-08 at 1 50 37 pm" src="https://user-images.githubusercontent.com/1711642/32571136-fd7516ba-c48b-11e7-950c-dcd9dc684cb3.png">

tm4027:
JIRA created

mjhenkes:
resolved with #225

PR #203, opened 2017-10-10
Make header compatible with the collapsible menu view

kschuste:
### Summary
This fixes issue #192.
There currently is no max-width property set on the start or end elements for the Header.  This allows the side elements to be 100% of their content.  When these elements are super big, this can cause any text in the middle to not be shown or even push the content off to the right.  Therefore, a max-width is needed for these areas.

I used a max-width of 40% for the side elements.  This will allow for the case when they are both extremely large, there will still be 20% of the content available for the text in the middle.

### Additional Details
This is how the the header works today with large side elements.
screen shot 2017-10-10 at 1 54 24 pm
screen shot 2017-10-10 at 1 47 38 pm

By using the max-width attribute, this is how it appears with this PR.
screen shot 2017-10-10 at 1 50 37 pm
screen shot 2017-10-10 at 1 48 51 pm


mjhenkes:
Is there anything we can do about the large button overflow issue?

kschuste:
@mjhenkes For the large button, that would be a separate issue to have the button text truncate rather than wrap at the white-space.  This is how our terra-buttons work today so the fix would need to be applied there.
screen shot 2017-10-10 at 3 21 05 pm

mjhenkes:
Another option would be allowing the header to grow.

kschuste:
@mjhenkes set a min-height on the header to allow for it to grow vertically if its content is larger.
screen shot 2017-10-10 at 4 58 33 pm

mjhenkes:
@kschuste make sure you bet this with ux. The implications deserve discussion.

neilpfeiffer:
Getting input from a few others in UX to make final determination.

neilpfeiffer:
Recap from yesterday - new requirements are to expand the height to show a full button if text wraps to a second (or more) lines, also new change is to convert the truncation of the title (center area) to always wrap instead of truncate, and height of header would grow as well to show full text. 

Still to be determined -- the appropriate widths of the regions (40% etc.) - still investigating and will work out an answer.

CONTRIBUTORS.md
```diff
@@ -15,3 +15,4 @@ Cerner Corporation
 [@tbiethman]: https://github.com/tbiethman
 [@mjhenkes]: https://github.com/mjhenkes
 [@nagrawal3]: https://github.com/nagrawal3
+[@kschuste]: https://github.com/kschuste
```
StephenEsser: You'll want to add your name above, your name doesn't show up on the view only mark down file

kschuste: done

packages/terra-clinical-header/tests/nightwatch/LargeLeftAndRightContentsWithLargeTitleHeader.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
```
emilyrohrbough: nit: exta space

packages/terra-clinical-header/tests/nightwatch/header-spec.js
```diff
@@ -50,4 +50,11 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
     browser.url(`${browser.launchUrl}/#/tests/header-tests/no-title-right-content`);
     browser.expect.element('#headerTest--content').to.be.present;
   },
+
+  'Displays a header with large left and right contents with a large title': (browser) => {
+    browser.url(`${browser.launchUrl}/#/tests/header-tests/large-left-and-right-contents-large-title`)
+    .assert.containsText('#Header h1', 'LongTitle');
```
emilyrohrbough: We want to use nightwatch's expect syntax here, not assert.

emilyrohrbough: I'm looking at this test, it seems validating it rendered would suffice given this is test example is to verify expanding height and we have already tested the header contains each piece in the test on line 31.

emilyrohrbough:
Is this what we would expect? 

screen shot 2017-10-26 at 2 44 08 pm

OR
screen shot 2017-10-26 at 2 45 29 pm

emilyrohrbough:
Should this test be updated to show height adjustment? Or is this needed anymore since we removed `no wrap` with `normal`?

https://terra-clinical-deployed-pr-203.herokuapp.com/static/#/tests/header-tests/long-title

kschuste:
@emilyrohrbough update the test to make it more pleasing to look at, the text wasn't wrapping at it was all one-word so was being truncated based off of word-wrap: normal rules; therefore, updated it to be break-word so even a really long word will wrap.
screen shot 2017-10-26 at 4 12 03 pm

packages/terra-clinical-header/tests/nightwatch/header-spec.js
yuderekyu: @emilyrohrbough What is the reason behind using expect, rather than assert?

packages/terra-clinical-header/src/Header.scss
```diff
@@ -46,10 +47,10 @@
...
     overflow: hidden;
     text-overflow: ellipsis;
-    white-space: nowrap;
+    white-space: normal;
     width: 100%;
-    word-break: normal;
-    word-wrap: normal;
+    word-break: break-word;
+    word-wrap: break-word;
   }
```
StephenEsser: `it cannot uncollapse the elements from the dropdown.`

This was taken from the pull request description above. Were you able to find a solution for this?

kschuste: Not without breaking other scenarios.  Resizing the window down and back bigger again is something that I feel is a rare enough use case that it is ok.

StephenEsser: I think a screen rotation could trigger this situation

kschuste: updated to handle the resizing by setting the widths on resize now

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -35,40 +37,142 @@ const defaultProps = {
...
-      <div className={cx('title-container')}>
-        <h1 className={cx('title')}>
-          {title}
-        </h1>
-      </div>
-    );
+  componentDidMount() {
+    // grab the initial full display widths of the end containers for reference
+    // when resizing the end containers inside handleOnResize
+    const startNode = ReactDOM.findDOMNode(this.startContainerTarget);
```
jmsv6d: You shouldn't need to use findDOMNode here. You set `this.startContainerTarget` as the dom node already in the ref callback.

kschuste: updated to not use findDOMNode

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -35,40 +35,138 @@ const defaultProps = {
...
-  let startElement;
-  if (startContent) {
-    startElement = <div className={cx('flex-end')}>{startContent}</div>;
+class Header extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleOnResize = this.handleOnResize.bind(this);
+    this.state = {
+      startContentInitialWidth: undefined,
+      endContentInitialWidth: undefined,
```
jmsv6d: I don't think these initial widths need to be stored in the state. State updates triggers the component to rerender and we shouldn't need to do that in order to set the initialWidth. Just set `this.startContentInitialWidth` and `this.endContentInitialWidth` in componentDidMount instead.

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -35,40 +35,138 @@ const defaultProps = {
...
+    const containerWidth = this.containerTarget.offsetWidth;
+    const fitMaxWidth = containerWidth * 0.4;
+
+    if (this.startContainerTarget) {
+      // set the start container to its initial width if can fit inside the 40% maxWidth
+      // else set the start container to the maximum possible width it can be
+      if (fitMaxWidth >= this.state.startContentInitialWidth) {
+        this.state.startContentWidth = this.state.startContentInitialWidth;
+      } else {
+        this.state.startContentWidth = fitMaxWidth;
```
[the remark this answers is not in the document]
kschuste: updated to use setState

packages/terra-clinical-header/src/Header.jsx
kschuste: removed these two items

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -35,40 +35,138 @@ const defaultProps = {
...
+    // grab the initial full display widths of the end containers for reference
+    // when resizing the end containers inside handleOnResize
+    if (this.startContainerTarget) {
+      this.startContentInitialWidth = this.startContainerTarget.offsetWidth;
+    }
+    if (this.endContainerTarget) {
+      this.endContentInitialWidth = this.endContainerTarget.offsetWidth;
+    }
+
+    window.addEventListener('resize', this.handleOnResize);
```
jmsv6d: Should we use the `resize-observer-pollyfill` that we're using in our other responsive components? Listening for the resize event on the window won't work when the header width changes because the parent container resizes. Ex. A slide panel that maximizes and minimizes.

jmsv6d:
It looks like the left and right content is no longer being vertically centered.
screen shot 2017-11-08 at 10 02 28 am

kschuste:
updated to vertically center the left and right contents, I think it looked like it was being centered vertically before, but we never allowed the header to expand vertically until now
screen shot 2017-11-08 at 10 37 07 am

kschuste:
Closing this PR until can find a 1 size fits all solution

Issue #216, opened 2017-10-26
Item Collection: Add subheaders to the ItemCollection


tbiethman:
# Issue Description
Some consumers have expressed the need to have subheaders within the ItemCollection. With the coming updates to the ItemCollection API, this might be more possible.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Subheaders can be rendered between items in the item collection (similar to the list/table implementations).

## Current Behavior
Subheaders cannot be added through the prop-based API.

emilyrohrbough:
@BenBoersma did you add an example to https://github.com/cerner/terra-clinical/pull/687 to show how to do this?

BenBoersma:
Currently there isn't example on how to do this, but I can easily put one in.

Issue #217, opened 2017-10-26
Item View: Add Props to Handle Missing Accessory Alignment


emilyrohrbough:
# Issue Description
Add additional props to add the appropriate spacing if an accessory is not present. 

For example, in the scenario that several Item Views with varying content are stacked underneath one anther, one would expect the displays to align, even if the content is not present. These props could be passed to handle this.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Allow ItemView to add the appropriate accessory spacing if there is no accessory and a bool prop is provided.
<img width="721" alt="screen shot 2017-10-26 at 9 16 39 am" src="https://user-images.githubusercontent.com/14099737/32057986-6eb639e8-ba2e-11e7-9f46-3e4b9527f5b9.png">

### Current Behavior 
<img width="737" alt="screen shot 2017-10-26 at 9 14 52 am" src="https://user-images.githubusercontent.com/14099737/32057904-30ff3866-ba2e-11e7-82dd-e352ad55ecdc.png">

PaulSolDev:
Orion JIRA created

tm4027:
JIRA created

PR #221, opened 2017-11-07
Item Collection: Enhance API

### LINKED ISSUE #213: Item Collection: Column span should be determined by the row with the most columns
# Issue Description

The ItemCollection column span is currently determined by the first row. The entire collection can only have as many columns as the first row has. Displays, accessories, and comments in subsequent rows are ignored if they exceed the column span of the first row.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

Row column spans are not tied to the column span of the first row

## Current Behavior

Each row is limited to the column span of the first row 

## Steps to Reproduce

1. Create an Item Collection similar the one displayed here: http://engineering.cerner.com/terra-clinical/#/site/item-collection
2. Add additional displays, comments, or accessories to any row other than the first
3. Observe the added content is not displayed if it exceeds the column span of the first row
( It is also possible to delete items in the first row and see the following row column span reduce )

## Environment

* Component Version:
* Browser Name and Version:
* Operating System and version (desktop or mobile):

### LINKED ISSUE #218: Item Collection: Enhance API
# Issue Description
The current prop `row` takes an array of hashes to generate the list and table displays. This is hard to use and inflexible for future enhancements. 

Tech Design Considerations: 
- Support Children API
- Support Headers and Subheaders #216 
- Options between static, single select and multi select rendering

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

### NOTE: This will be a major version bump.

emilyrohrbough:
Closes #218 

#### Props Added
- `children` - The Item Collection Items that will flex between either a list item or a table row.
- `requiredElements` - the elements expected to be displayed. If a child is missing an element, ItemCollection will allocate space to maintain the provided layout.
- `onSelect` - the callback function that is assigned as to a child's onClick and onKeyDown callback if the child isSelectable. The first parameter returned is the event and the second parameter is the child key.

#### Props Changed
- `rows` - replaced with `children` prop
- `onChange` - replaced with `onSelect` prop
- `listStyles` - removed `hasChevrons` key. Chevrons are intended for single-selectable lists only to indicate clicking the item will reveal more information. This behavior is not currently present in selectable tables.

#### Added/Changed Behavior
* Allows for either static and selectable implementation
* OnSelect returns the event and child key for onClicks and onKeyDowns
* Exports the following components: ItemCollection.Item, ItemCollection.Display and ItemCollection.Comment.
* Table layout no longer dependent on first Item Collection item - Closes #213

## Additional Notes
* Set up to allow for Header and Subheader support - #216
* Removed `_TableHeaderCell` for a different column width implementation.

## Deployment
- https://terra-clinical-deployed-pr-221.herokuapp.com/static/#/site/item-collection
- https://terra-clinical-deployed-pr-221.herokuapp.com/static/#/tests/item-collection-tests
## ~Coming Soon~ Done
- [x] Selectable Implementation Guide
- [x] [V1 to V2 Migration Guide](https://gist.github.com/emilyrohrbough/825faffe56641b19b5071d339adbc1f8)


packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,160 @@
...
+
+const propTypes = {
+  /**
+   * The display elements to be presented.
+   */
+  children: PropTypes.node,
+  /**
+   * The start accessory element to be presented.
+   */
+  startAccessory: PropTypes.node,
```
emilyrohrbough: `startAccessory`, `comment`, and `endAccessory` should be of type element in my option to reflect the ItemView API. The reason I chose type node was that the [TableCell component does not allow empty content](https://github.com/cerner/terra-core/blob/master/packages/terra-table/src/TableCell.jsx#L9). When creating the table displays, this meant I would need to supply the content with a filler element or an empty string. An empty string seemed more appropriate & type node allows for this string, but resulted in wrapping the `startAccessory`, `comment`, and `endAccessory` in a div to ensure they are an element.

@bjankord @mjhenkes Do we remember why TableCell requires content? Given the component type, it seems there was a good reason since I'd expected empty cells in a table.

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,160 @@
...
+      onSelect(event, itemKey);
+    };
+  }
+
+  return selectableProps;
+}
+
+function createListItem(elements, itemKey, selectableProps, isSelected, listItemStyles) {
+  const { startAccessory, children, comment, endAccessory, hasStartAccessory, hasEndAccessory, ...listItemProps } = elements;
+  // hasStartAccessorySpacing={hasStartAccessory}
```
emilyrohrbough: This will be hooked into #217 once completed. May or may not be the prop names we'd expect but wanted it set up.

packages/terra-clinical-item-collection/src/_CreateListView.jsx
```diff
@@ -1,50 +0,0 @@
-import React from 'react';
```
emilyrohrbough: Replaced with `_ListView`. Implementation is quite different.

packages/terra-clinical-item-collection/src/_CreateTableView.jsx
```diff
@@ -1,121 +0,0 @@
-import React from 'react';
```
emilyrohrbough: Replaced with `_TableView`. Implementation is quite different.

packages/terra-clinical-item-collection/src/_DetermineElements.js
```diff
@@ -1,15 +0,0 @@
-const determineElements = (row) => {
```
emilyrohrbough: This was utilized by `_TableHeaderCell` which was removed for a different column width implementation.

packages/terra-clinical-item-collection/src/_TableHeaderCell.jsx
```diff
@@ -1,29 +0,0 @@
-import React from 'react';
```
emilyrohrbough: Removed for a different column width implementation.

packages/terra-clinical-item-collection/tests/nightwatch/item-collection/ItemCollectionTests.jsx
```diff
@@ -7,19 +7,14 @@ const ItemCollectionTests = () => (
   <div>
     <ul>
       <li><Link to="/tests/item-collection-tests/default">Item Collection - Default</Link></li>
-      <li><Link to="/tests/item-collection-tests/tiny-breakpoint">Item Collection - Tiny Breakpoint</Link></li>
```
emilyrohrbough: These breakpoint tests were testing that the responsive element worked correctly and weren't needed.

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,160 @@
...
+    console.error('Key is required for correct selectable implementation.');
+    /* eslint-disable no-console */
+  }
+
+  if (onSelect) {
+    selectableProps.onClick = (event) => {
+      onSelect(event, itemKey);
+    };
+
+    selectableProps.onKeyDown = (event) => {
```
emilyrohrbough: Should we handle the keydown selection to be consistent with List & Table?

```
if(event.nativeEvent.keyCode === KEYCODES.ENTER || event.nativeEvent.keyCode === KEYCODES.SPACE) {
onSelect(event, itemKey);
}
```

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,160 @@
...
+
+function createTableCell(content, keyValue, contentType) {
+  return (<Table.Cell content={content} key={keyValue} className={cx(`content-${contentType}`)} />);
+}
+
+function createTableRow(elements, itemKey, selectableProps, isSelected) {
+  const { startAccessory, children, comment, endAccessory, ...tableRowProps } = elements;
+
+// QUESTION:
+  // Does ITEM as a standalone component want to care about the maximun number of displays columns?
```
kschuste: I would expect ITEM to be agnostic and not care what it is being placed inside and let the container handle that logic

packages/terra-clinical-item-collection/src/ItemCollection.jsx
```diff
@@ -2,123 +2,102 @@ import React from 'react';
...
   breakpoint: PropTypes.oneOf(['tiny', 'small', 'medium', 'large', 'huge']),
   /**
-   * The styles to spread to the table. Table style options are isPadded and isStriped.
+   * The items that will flex between either a list item or a table row.
    */
-  tableStyles: PropTypes.shape({
-    isPadded: PropTypes.bool,
-    isStriped: PropTypes.bool,
+  children: PropTypes.node.isRequired,
+  /**
```
emilyrohrbough: TODO: Flatten theses to `hasStartAccessory`, `numberOfDisplays`, `hasComment`, `hasEndAccessory`

@StephenEsser @tbiethman

packages/terra-clinical-item-collection/src/ItemCollection.jsx
```diff
@@ -2,123 +2,102 @@ import React from 'react';
...
-    itemStyles: PropTypes.shape({
-      layout: PropTypes.oneOf(['oneColumn', 'twoColumns']),
-      textEmphasis: PropTypes.oneOf(['default', 'start']),
-      isTruncated: PropTypes.bool,
-      accessoryAlignment: PropTypes.oneOf(['alignTop', 'alignCenter']),
-    }),
-  })),
-  /**
-   * A callback event that will be triggered when selection state changes.
+   * The styles to spread to the table. Table style options are isPadded and isStriped.
```
emilyrohrbough: TODO: flatten to `isTablePadded` and `isTableStriped`

packages/terra-clinical-item-collection/src/ItemCollection.jsx
```diff
@@ -2,123 +2,102 @@ import React from 'react';
...
   }),
   /**
-   * The styles to spread to the list. List style options are isDivided and hasChevrons.
+   * The callback function that is assigned as to a child's onClick and onKeyDown callback if the child isSelectable. The first
+   * parameter returned is the event and the second parameter is the child key. Function is not applied if child is not selectable.
+   */
+  onSelect: PropTypes.func,
+  /**
+   * The styles to spread to the list. List style options are isDivided.
    */
```
emilyrohrbough: TODO: flatten to `isListDivided`

packages/terra-clinical-site/src/examples/item-collection/ItemCollectionExample.jsx
```diff
@@ -1,58 +1,88 @@
 import React from 'react';
```
emilyrohrbough: TODO: Expose ItemCollection.Item Props & Use.

packages/terra-clinical-item-collection/CHANGELOG.md
```diff
@@ -3,6 +3,18 @@ ChangeLog
...
+
+### Added
+* onSelect, children, and requiredElements props
+* Item Component to use as for Item Collection children
+
+### Removed
+* rows and onChange props
+
+### Changed
+* Item Collection layout is no longer rendered via the first row, but the requiredElements object
```
StephenEsser: There are several references to requiredElements that may not make contextual sense after flattening the props

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,189 @@
...
+  view: 'list',
+};
+
+function createSelectableProps(onSelect, itemKey) {
+  const selectableProps = { isSelectable: true };
+  selectableProps.tabIndex = 0;
+
+  if (!itemKey) {
+    /* eslint-disable no-console */
+    console.error('Key is required for correct selectable implementation.');
```
kschuste: seems like we should set and return isSelectable: false and tabIndex: -1 here right away then if no key and it is required for correct implementation

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,189 @@
...
+      }
+    };
+  }
+
+  return selectableProps;
+}
+
+function createListItem(elements, itemKey, selectableProps, isSelected, listItemStyles) {
+  const { startAccessory, children, comment, endAccessory, hasStartAccessory, hasEndAccessory, ...listItemProps } = elements;
+  // alignStartAccessory={hasStartAccessory}
```
kschuste: extra commented lines here not needed

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,189 @@
...
+  return (<Table.Cell content={content} key={keyValue} className={cellClassNames} />);
+}
+
+function createTableRow(elements, itemKey, selectableProps, isSelected, accessoryAlignment) {
+  const { startAccessory, children, comment, endAccessory, ...tableRowProps } = elements;
+
+// QUESTION:
+  // Does ITEM as a standalone component want to care about the maximun number of displays columns?
+  // The max is 8 - determined by ItemView rendering.
+  // When used in the ItemCollection, _createTableView handles the case of too many.
```
kschuste: my vote is No, the container should care, not the individual items

packages/terra-clinical-item-collection/src/Item.jsx
emilyrohrbough: Updating ItemView's props to accept nodes in #222

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
```diff
@@ -0,0 +1,89 @@
...
+ * @ param {childElement} obj - The react child props that will compose a table row.
+ * @ param {requiredElements} obj - The elements expected to be displayed.
+ */
+function addAnyMissingTableElements(childElements, requiredElements) {
+  const { hasStartAccessory, numberOfDisplays, hasComment, hasEndAccessory } = requiredElements;
+  const { startAccessory, children, comment, endAccessory } = childElements;
+  const tableRowPieces = { startAccessory: null, children: null, comment: null, endAccessory: null };
+
+  if (hasStartAccessory) {
+    tableRowPieces.startAccessory = ' ';
```
kschuste: could put this in an else bracket to not have to set this twice

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
```diff
@@ -0,0 +1,89 @@
...
+      displayContent = children.length ? children.slice(0, numberOfDisplays) : [children];
+    }
+    while (displayContent.length < numberOfDisplays) {
+      displayContent.push(' ');
+    }
+  }
+  tableRowPieces.children = displayContent;
+
+  if (hasComment) {
+    tableRowPieces.comment = ' ';
```
kschuste: can put this in an else bracket to not have to write to comment twice

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
```diff
@@ -0,0 +1,89 @@
...
+    if (comment) {
+      tableRowPieces.comment = comment;
+    }
+  }
+
+  if (hasEndAccessory) {
+    tableRowPieces.endAccessory = ' ';
```
kschuste: can put this in an else bracket to not have to write to endAccessory twice

packages/terra-clinical-item-collection/tests/nightwatch/item-collection/clinical-item-collection-spec.js
```diff
@@ -3,71 +3,50 @@
...
+    browser.click('#item1');
+    browser.expect.element('#selected-key').text.to.contain('1');
   },
 
-  'Displays the correct collection responsive to window size for huge breakpoint': (browser) => {
-    browser.url(`${browser.launchUrl}/#/tests/item-collection-tests/huge-breakpoint`);
-    browser.execute("document.getElementById('siteApplication').style.padding = 0;");
+  'Returns child key for ENTER when onSelect is given to a selectable item collection': (browser) => {
+    browser.url(`${browser.launchUrl}/#/tests/item-collection-tests/selectable`);
+    browser.expect.element('#item3').to.be.present;
```
kschuste: if added a `.to.equal(true)` here can remove the ignore of the linter rule of no-unused-expressions

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,189 @@
...
+    isSelected,
+    onSelect,
+    listItemLayout,
+    listItemTextEmphasis,
+    isListItemTruncated,
+    accessoryAlignment,
+    ...otherItemProps
+  } = props;
+  // The itemKey is set by the ListView & TableView components as the child.key value.
+  const { itemKey, ...elements } = otherItemProps;
```
jmsv6d: So... this seems questionable to me. You using custom props in order to send the key, I assume so that this prop is not visible to consumers. Then you throw an error if we don't have it, essentially making it a required prop, but it's not documented anywhere except the console error.

packages/terra-clinical-item-collection/src/Item.jsx
emilyrohrbough: This only throws an error & will still create the onClick and onKeyDown events for the item. However the key is the callback for these function, so if key isn't specified when the ItemCollection was created, this will return null.

packages/terra-clinical-item-collection/src/Item.jsx
emilyrohrbough: So I 100% agree this seems very odd. React [deprecated the ability for a component to access its own key](https://reactjs.org/blog/2014/10/16/react-v0.12-rc1.html#breaking-change-key-and-ref-removed-from-thisprops) a while back. The response was to create a different 'read-only' prop to pass the key down to the component. 

The ItemCollection handles this mapping to the itemKey 'read-only' prop [here](https://github.com/cerner/terra-clinical/pull/221/files#diff-fce8e1aeaf3a85769a01bcfc4c3bd946R61) and [here](https://github.com/cerner/terra-clinical/pull/221/files#diff-a8fa6bc226c2f3aa668a0e94d6f34228R36). 

Because we'd expect people to be using keys to manager the children, I didn't expose the itemKey prop because a consumer wouldn't need to define the itemKey & key props. They would only care about itemKey if they used Item outside of the ItemCollection.

I could do a couple things:
1. Expose the ItemKey prop with documentation regarding it should only be used when Item is not placed in the ItemCollection
2. Expose the prop, honor itemKey over key

I feel option two could becoming more overhead than its worth. I do thinking option one might be okay since @StephenEsser mention people are looking to create list-only or table-only collection and this would give them them the structor for consistent styling.

I did include an error console message if key was not provided in createSelectableProps() to indicate if key wasn't provided. I found that React requires keys/unique keys, it will accept null key values. 

@cerner/terra thoughts on this?

kschuste: I like option 1, would there be any issues if providing this prop if used in item collection?

emilyrohrbough: If provided, it would be overwritten by the key value provided.

kschuste: I'm still ok with option 1 then

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * When displayed as a list item, whether or not all text should truncate.
+   */
+  isListItemTruncated: PropTypes.bool,
+  /**
+   * The vertical alignment of the start and end accessories.
+   */
+  accessoryAlignment: PropTypes.oneOf(['alignTop', 'alignCenter']),
+  /**
+   * Wether or not the item is selectable. If true, the item is given list and table hover and focus styles, tabIndex
```
JakeLaCombe: *Whether.

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,188 @@
...
+};
+
+function createSelectableProps(onSelect, itemKey) {
+  const selectableProps = { isSelectable: true };
+  selectableProps.tabIndex = 0;
+
+  if (onSelect) {
+    if (!itemKey) {
+      /* eslint-disable no-console */
+      console.error('Key is required for correct selectable implementation.');
```
JakeLaCombe: If they pass an onSelect without an itemKey, do we expect the component to error out, or just ignore the requested implementation as it's invalid? If it's the later, it maybe better to do console.warn inside a non production environment by using the following wrapper

```
if (process.env !== 'production') {
}
```

If it's the first one where we expect an error, should we also do something when onSelect isn't valid as well?

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,188 @@
...
+
+function createTableRow(elements, itemKey, selectableProps, isSelected, accessoryAlignment) {
+  const { startAccessory, children, comment, endAccessory, ...tableRowProps } = elements;
+
+// QUESTION:
+  // Does ITEM as a standalone component want to care about the maximun number of displays columns?
+  // The max is 8 - determined by ItemView rendering.
+  // When used in the ItemCollection, _createTableView handles the case of too many.
+
+  let displayContent = [];
```
JakeLaCombe: I think this can be combined into one line, as map always returns an array

```
let displayContent = React.Children.map(...
```

packages/terra-clinical-item-collection/src/Item.jsx
JakeLaCombe: I vote for 1 as well

packages/terra-clinical-item-collection/src/ItemCollection.jsx
```diff
@@ -2,123 +2,133 @@ import React from 'react';
...
+  /**
+  * Whether or not the item collection should create a layout to display a start accessory.
+  */
+  hasStartAccessory: PropTypes.bool,
+  /**
+  * The number of displays the item collection should create a layout for.
+  */
+  numberOfDisplays: PropTypes.number,
+  /**
+  * Whether or not the item collection should create a layout to display a comment.
```
JakeLaCombe: I feel these documentation comments are too implementation specific, and a bit confusing to the consumers. Instead of saying create a layout, I think it's just fine to say 'where or not the item collection contains a comment'.

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
```diff
@@ -0,0 +1,79 @@
...
+};
+
+/**
+ * This function ensures the correct elements are provided to create a consistent item view layout. To handle a
+ * potential missing required accessory element, a boolean indicator is provided to single accessory space is needed.
+ *
+ * @ param {childElement} obj - The react child props that will compose an item view list item.
+ * @ param {requiredElements} obj - The elements expected to be displayed.
+ */
+function addAnyMissingListElements(childElements, requiredElements) {
```
JakeLaCombe: Nitpick:

I think Any is extraneous here. addMissingListElements seems sufficient enough imo

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
```diff
@@ -0,0 +1,79 @@
+const KEYCODES = {
+  ENTER: 13,
+  SPACE: 32,
+};
+
+/**
+ * This function ensures the correct elements are provided to create a consistent item view layout. To handle a
+ * potential missing required accessory element, a boolean indicator is provided to single accessory space is needed.
+ *
+ * @ param {childElement} obj - The react child props that will compose an item view list item.
```
JakeLaCombe: It sounds like these are childProps and parentProps if I'm reading this correctly?

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,188 @@
...
+   * The vertical alignment of the start and end accessories.
+   */
+  accessoryAlignment: PropTypes.oneOf(['alignTop', 'alignCenter']),
+  /**
+   * Wether or not the item is selectable. If true, the item is given list and table hover and focus styles, tabIndex
+   * set to 0, and onClick and onKeyDown callbacks set to the onSelect function provided via Item Collection.
+   */
+  isSelectable: PropTypes.bool,
+  /*
+   * Wether or not the item is selected.
```
kschuste: *Whether

packages/terra-clinical-item-collection/src/ItemCollection.jsx
```diff
@@ -2,123 +2,133 @@ import React from 'react';
...
+  */
+  hasComment: PropTypes.bool,
+  /**
+  * Whether or not the item collection should create a layout to display an end accessory.
+  */
+  hasEndAccessory: PropTypes.bool,
+  /**
+   * The callback function that is assigned as to a child's onClick and onKeyDown callback if the child isSelectable. The first
```
kschuste: change isSelectable to is selectable

packages/terra-clinical-item-collection/src/_ListView.jsx
```diff
@@ -0,0 +1,55 @@
...
+      const itemViewPieces = Utils.addAnyMissingListElements(child.props, requiredElements);
+      return React.cloneElement(child, { view: 'list', itemKey: child.key, onSelect, ...itemViewPieces });
+    }
+
+    return child;
+  });
+}
+
+const ListView = ({ children, onSelect, isListDivided, requiredElements }) => {
+  const listItems = createListItems(children, onSelect, requiredElements);
```
JakeLaCombe: I think you can leave out the listItems variable all together, making this a function not needing curly braces

packages/terra-clinical-item-collection/src/_TableView.jsx
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * The items be rendered as a table row.
+   */
+  children: PropTypes.node,
+  /**
+   * The callback function that is assigned as to a child's onClick and onKeyDown callback if the child isSelectable.
+   */
+  onSelect: PropTypes.func,
+  /**
+   * Whether or not the table rows should be zebra striped.
```
JakeLaCombe: Looks like the documentation on props are switched

packages/terra-clinical-item-collection/tests/jest/Item.test.jsx
```diff
@@ -0,0 +1,335 @@
...
+        comment={<Item.Comment text="comment" />}
+        endAccessory={<p>end</p>}
+        isSelectable
+      >
+        <Item.Display text="Display 1" />
+      </Item>
+    );
+    shallow(item);
+  } catch (e) {
+    expect(e.message).toContain('Key is required for correct selectable implementation.');
```
JakeLaCombe: Is this catching the exception? I was thinking you need

```
throw new Error('oops');
```

as opposed to `console.error` to trigger this.

packages/terra-clinical-item-collection/tests/nightwatch/item-collection/ItemCollectionSelectable.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+  handleSelection(event, selectedItemKey) {
+    this.setState({ selectedItemKey });
+  }
+
+  render() {
+    return (
+      <div>
+        <h3>This a POC Selectable ItemCollection Example.</h3>
+        <p>This example it set up to be a sinlge-selection example, however the Item Colleciton
```
JakeLaCombe: Couple of typos here.

packages/terra-clinical-item-collection/tests/nightwatch/item-collection/clinical-item-collection-spec.js
JakeLaCombe: To be honest, I think the format that it's in here is more readable, and that adding that boolean check could make the code a little more confusing. I would rather just disable the linter here, as we have a good reason for doing it.

JakeLaCombe:
Can a selectable example be added to the examples page?

emilyrohrbough:
@JakeLaCombe I do not want to add an selectable example to the example page because the selectable implementation in 100% up to the consumer & I would not want to give the impression any certain behavior should be expected by the item collection. I added the test page to ensure it functioned as expected. If others also feel I should add an example to the site page, I can do so.

packages/terra-clinical-item-collection/src/ItemCollection.scss
```diff
@@ -1,10 +1,31 @@
...
+    table-layout: fixed;
+  }
+
+  .start-accessory-column {
     width: 40px;
   }
 
-  .content-accessory {
+  .end-accessory-column {
+    width: calc(100% / 7);
```
tbiethman: I talked offline with @emilyrohrbough regarding this width for the end accessory column. It provides the end accessory column with more real estate than I'd expect in the examples, and I'm not sure we can land on a fixed value here that will satisfy the majority of implementations.

Emily suggested giving the end accessory the same width as the other text/comment columns, which I think makes sense. If the end accessory column width ever needs to be customizable or dynamic, we should investigate adding custom column width support across the entire table like we've talked about before. Until then, I think it makes sense to be as un-opinionated as we can be.

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
emilyrohrbough: Could you expand on your statement? I'm not sure what you mean by parent props. But I from what I think you are asking, 

requriedElements is an object that passes the `hasStartAccessory`, `numberOfDisplays`, `hasComment` and `hadEndAccessory` prop values of the Item Collection component

and 

childElements are the element pieces (start/end accessory, comment and displays) of each ItemCollection Item.

packages/terra-clinical-item-collection/src/Item.jsx
emilyrohrbough: @jmsv6d What are your thoughts?

packages/terra-clinical-item-collection/src/Item.jsx
emilyrohrbough: Consumers should be providing an 'key' on each item that will be mapped to the 'read-only' itemKey prop if they are implementing a selectable Item Collection. React allows will catch duplicate key values, however a key does not have to be provided or could can be null. 

I added this catch for the instance a consumer did not provide a key because the Item Collection will render as expected with the correct selectable behaviors & styling, however the catch is the itemKeys callback will be undefined. IMO the component should not error out because the lack of a valid key is an implementation issue on the consumer's side.

packages/terra-clinical-item-collection/src/Item.jsx
jmsv6d: I'm okay with option 1, but it wouldn't be my first choice. As is, I definitely think that documentation would be needed, especially if people will be wanting to use this component outside of the item collection. However, I think that some refactoring could be done to remove the need for it at all. It's really only needed as a parameter for the onSelect function. If that logic was lifted up to a parent component I don't think that you'd need the 'read-only' prop at all.

emilyrohrbough: hmmm okay. I see what you are saying. Good call. I made some changes in [this commit](https://github.com/cerner/terra-clinical/pull/221/commits/7d8429c76e241a9157085530f2459bb32d290a53). Let me know if this is what you were thinking.

This commit moves the onSelect event allocation to the Utils files while keeping the isSelectable, tabIndex and isSelected logic within the Item file. Then for creating a selectable Item Collection there are no changes in implementation, however, anyone wanting to use Item outside of the IC can still receive tabIndex, isSelected and selectable styles but would need to attach their own onClick and onKeyDown functions for full selectability.

packages/terra-clinical-item-collection/src/Item.jsx
JakeLaCombe: I'm fine with that. In that case, can you switch to a console.warn?

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+/**
+ * This function ensures the correct elements are provided to create a consistent item view layout. To handle a
+ * potential missing required accessory element, a boolean is provided to indicate if start accessory space is needed.
+ *
+ * @ param {childElement} obj - The react child props that will compose an item view list item.
+ * @ param {requiredElements} obj - The elements expected to be displayed.
+ */
+function addMissingListElements(childElements, requiredElements) {
```
JakeLaCombe: I'll take another stab at this! I know I was confusing last time!

In the use of this method (https://github.com/cerner/terra-clinical/pull/221/files#diff-a8fa6bc226c2f3aa668a0e94d6f34228R35), I see that childElements corresponds to the props of the child, and requiredElements corresponds to props of the ListView. Given that, I think these should be renamed to childProps and requiredProps. Personally though, I think requiredProps should be parentProps, as it looks like it's taking props from the parent ListView.

emilyrohrbough: ^ I see what you are saying here. I can update these to be childProps.

I also see what you mean about `requiredElements` as well. I wanted the concept of the requiredElements obj as these are the pieces needed to create a consistent layout. I am open to a different name, but parentProps to me would be more confusing and not a true name to represent what is being passed. When working through the code as requiredProps are created by the Item Collection and the same object is passed to the _tableView and _ListView and both component pass requiredElements to Utils.

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+const Item = (props) => {
+  const {
+    view,
+    isSelectable,
+    isSelected,
+    listItemLayout,
+    listItemTextEmphasis,
+    isListItemTruncated,
+    accessoryAlignment,
```
kschuste: nil:  could list these in alphabetical order to match up with the declaration of them at the top

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
kschuste: the documentation says the first param is props and the next contains elements, so we should just change the first params name to childProps

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/**
+ * This function ensures the correct elements are provided to create a consistent table row layout. If a required
+ * element is missing, an empty div is provided as cell content.
+ *
+ * @ param {childElement} obj - The react child props that will compose a table row.
+ * @ param {requiredElements} obj - The elements expected to be displayed.
+ */
+function addMissingTableElements(childElements, requiredElements) {
```
kschuste: looks like the same as above

packages/terra-clinical-item-collection/src/_ItemCollectionUtils.jsx
JakeLaCombe: I see what requiredElements are. It maps directly to the prop types of the listView. I like the structure it's going for with requiredElements, but I don't know if element is the best word. When I think element, I think of an html node or react component. For that prop, perhaps it's better to say requiredProps or requiredAttributes?

JakeLaCombe: I see what requiredElements are now. In that case, I would recommend childProps and requiredElements.

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,151 @@
...
+   * When displayed as a list item, whether or not all text should truncate.
+   */
+  isListItemTruncated: PropTypes.bool,
+  /**
+   * The vertical alignment of the start and end accessories.
+   */
+  accessoryAlignment: PropTypes.oneOf(['alignTop', 'alignCenter']),
+  /**
+   * Whether or not the item is selectable. If true, the item is given list and table hover and focus styles, tabIndex
+   * set to 0, and onClick and onKeyDown callbacks set to the onSelect function provided via Item Collection.
```
jmsv6d: You are no longer setting the onClick and onKeyDown callbacks here.

packages/terra-clinical-item-collection/src/ItemCollection.scss
```diff
@@ -1,10 +1,27 @@
...
+    table-layout: fixed;
+  }
+
+  .start-accessory-column {
     width: 40px;
   }
 
-  .content-accessory {
+  .content-accessory-align-top {
+    vertical-align: top !important;
```
jmsv6d: Why are these `!important` needed?

emilyrohrbough: For whatever reason the terra-table styles are applied over the Item Collection styles

packages/terra-clinical-item-collection/src/_ListView.jsx
```diff
@@ -0,0 +1,56 @@
...
+import Item from './Item';
+import Utils from './_ItemCollectionUtils';
+
+const propTypes = {
+  /**
+   * The items be rendered as a list item.
+   */
+  children: PropTypes.node,
+  /**
+   * The callback function that is assigned as to a child's onClick and onKeyDown callback if the child isSelectable.
```
kschuste: nit: change isSelectable to is selectable.

packages/terra-clinical-item-collection/src/Item.jsx
```diff
@@ -0,0 +1,151 @@
...
+const Item = (props) => {
+  const {
+    accessoryAlignment,
+    isListItemTruncated,
+    isSelectable,
+    isSelected,
+    listItemLayout,
+    listItemTextEmphasis,
+    view,
+    ...elements
```
StephenEsser: Is `elements` here representing `customProps` or `children`? I think we should still spread custom props onto the TableRow / ItemView, or at the minimum allow an onClick to be passed through. 

Each ItemCollection.Item should still be able to respond to an onClick, is that possible with the current implementation?

emilyrohrbough: elements includes children, startAccessory, displays, endAccessory and customProps. Custom props are broken out [here](https://github.com/cerner/terra-clinical/pull/221/files/53645d2119839ea08d1ec8912d7aec688a7f305d#diff-93e672701138c7e816ff34f0d5b7dc53R74) & [here](https://github.com/cerner/terra-clinical/pull/221/files/53645d2119839ea08d1ec8912d7aec688a7f305d#diff-93e672701138c7e816ff34f0d5b7dc53R108), so custom props are spread on the list items and table rows. This should probably be cleaned up to make this clearer.

packages/terra-clinical-item-collection/src/ItemCollection.scss
jmsv6d: It looks like the table styles do have a higher specificity (1 class and 1 element). I think if you change these selectors to be `.content-accessory.content-accessory-align-top` then the specificity should be high enough with 2 classes to override the table styles without the `!important`

tm4027:
testing Item Collection in firefox, using the Item Collection - Varying Items Example  and Item Collection - Selectable Example tests, at the phone size the right accessory button is getting its left edge cut off.
terra-clinica pull 221

packages/terra-clinical-item-collection/src/ItemCollection.scss
emilyrohrbough: That worked! Thank you!!

emilyrohrbough:
@tm4027 This appears to be an issue with ItemView & overflow. Each browser handles this differently. I did some testing with https://terra-clinical-deployed-pr-221.herokuapp.com/static/#/tests/item-view-tests/accessory and it appears this displays incorrectly with Firefox, IE 10, Edge.

IE 10: 

screen shot 2017-11-29 at 11 35 34 am

EDGE:
screen shot 2017-11-29 at 11 35 22 am

PaulSolDev:
Functional review complete.  As explained by @emilyrohrbough the issue identified by @tm4027 is actually an problem in the item view.  No further issues identified.

Issue #115, opened 2017-06-28
Item-View, icons do not scale


Matt-Butler:
# Issue Description

At larger font-sizes, icons get cropped off at the bottom. 

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior

Icons scaled with font-size.

## Current Behavior

Icons are not scaled with font-size.
![screen shot 2017-06-28 at 12 44 08 pm](https://user-images.githubusercontent.com/8146159/27652003-c4fd88d4-5bff-11e7-9141-9be9bd2a4b08.png)

The person, comment, and information icon are cropped off at the bottom. 

## Steps to Reproduce

1. Open the item-view examples
2. Adjust the browser font-size to very large.

## Environment

* Component Version: 0.2.0
* Browser Name and Version: Chrome 
* Operating System and version (desktop or mobile): v58

emilyrohrbough:
> The person, comment, and information icon are cropped off at the bottom.

PR #156 should fix the current behavior seen with the person and comment being cut off.

Issue #114, opened 2017-06-28
Item-display, icon does not scale


Matt-Butler:
# Issue Description

![item-display](https://user-images.githubusercontent.com/8146159/27647375-e0eab698-5bf0-11e7-9356-f5aafca4625a.gif)

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior

Icon scales with font-size

## Current Behavior

Icon does not scale with font-size. At large font-sizes, icon is no longer visible.

## Steps to Reproduce

1. Navigate to the item display examples 
2. Increase or decrease the font-size of the item-display.

## Environment

* Component Version: 0.2.0
* Browser Name and Version: Chrome 
* Operating System and version (desktop or mobile): Desktop

PR #222, opened 2017-11-09
Item View: Add Props to Handle Accessory Alignment

### LINKED ISSUE #217: Item View: Add Props to Handle Missing Accessory Alignment

emilyrohrbough:
### Summary
Added `alignStartAccessory`, `alignEndAccessory` and `accessoryScale` props so the Item View can allocate the appropriate spacing if an accessory is not present but spacing is desired. 

Closes #217

### Deployment Link
https://terra-clinical-deployed-pr-222.herokuapp.com/static/#/tests/item-view-tests/accessory

Additionally, updated nightwatch to use expect syntax.

@cerner/terra


packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -47,26 +59,34 @@ const defaultProps = {
   textEmphasis: 'default',
   isTruncated: false,
   accessoryAlignment: 'alignCenter',
+  accessoryScale: '1em',
```
JakeLaCombe: If we use em as the default unit when the max is in px, is there a chance for large font sizes that this could be greater than the max scale?

emilyrohrbough: Does this last example answer your question?

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -25,48 +25,68 @@ const propTypes = {
    */
   accessoryAlignment: PropTypes.oneOf(['alignTop', 'alignCenter']),
   /**
+   * The height and width value the accessory should receive. Maximun scalar is 60px.
```
jmsv6d: Typo: Maximum

packages/terra-clinical-item-view/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* Added alignStartAccessory, alignEndAccessory and accessoryScale props to allow for proper spacing when a accessory is not provided.
```
tbiethman: nitpicking: ...for proper spacing when 'an' accessory is not provided.

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -25,48 +25,68 @@ const propTypes = {
...
+   * The height and width value the accessory should receive. Maximum scalar is 60px.
+   */
+  accessoryScale: PropTypes.string,
+  /**
    * The react element to be placed in the start aligned accessory position.
    */
-  startAccessory: PropTypes.element,
+  startAccessory: PropTypes.node,
+  /**
+   * Indicates whether or not the allow space for the start accessory if none is given.
```
tbiethman: I think you mean "to allow space" vs. "the allow space"

Same thing below for alignEndAccessory.

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -25,48 +25,68 @@ const propTypes = {
...
+  alignStartAccessory: PropTypes.bool,
   /**
    * The react element to be placed in the end aligned accessory position.
    */
-  endAccessory: PropTypes.element,
+  endAccessory: PropTypes.node,
+  /**
+   * Indicates whether or not space is allocated for the end accessory if none is given.
+   */
+  alignEndAccessory: PropTypes.bool,
```
StephenEsser: It might be late in the game for a change but these prop names match the props in Arrange but mean completely different things which is a little confusing. 

https://github.com/cerner/terra-core/blob/master/packages/terra-arrange/src/Arrange.jsx#L26-L34

The name implies the prop is performing some alignment to the start / end accessory rather than preserving or providing default spacing

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -25,48 +25,68 @@ const propTypes = {
...
+    if (accessory || alignAccessory) {
       const accessoryClassNames = cx(
         'accessory',
         { 'accessory-align-center': accessoryAlignment === 'alignCenter' },
         { 'accessory-align-top': accessoryAlignment === 'alignTop' },
       );
 
+      let scaledAccessory = accessory;
+      if (accessory && accessoryScale !== null) {
+        scaledAccessory = React.cloneElement(accessory, { width: accessoryScale, height: accessoryScale });
```
StephenEsser: This will spread width and height directly onto any accessory component. If the component isn't an Icon it will throw warning / errors on the dom for invalid props and override valid props if they are given.

Should the props be set directly on the icon accessory rather than passed through the ItemView to override the props on the Icon?

packages/terra-clinical-item-view/src/ItemView.jsx
emilyrohrbough: That is a good point. Originally I had thought something like `hasStartAccessorySpacing` but that was long & ugly. Any naming ideas?

packages/terra-clinical-item-view/src/ItemView.jsx
emilyrohrbough: Merging the existing props with the new props of `width` and `height` doesn't actually thrown warnings or errors on the dom for invalid props. React treats these as style props and will apply them as so. 

We could handle setting the accessory width, which is done [here](https://github.com/cerner/terra-clinical/pull/222/files#diff-8788aebd203a6b2eaee49ee3dab4e2e6R88).

The reason we have this logic scaling logic comes from these two issues logged a while back #114 & #115. The maximum height & width keeps the Icon scaled and positioned correctly within the space provided by the accessory when font-size scaled up.

StephenEsser: It does look like these are special props that are silently spread, but I tested spreading these onto React nodes and it didn't actually affect the styled width or height of the Component, are you able to see changes in width in height on any component other than Icon?

For the scaling, the issues suggest the icons should scale by only adjusting the font-size.  Should these Icons work similar to that of the List chevron which will adjust only by setting the font-size?

emilyrohrbough: hmmm I see your point regarding the scaling & it being an icon specific implementation. & Yes I do think icons should only be adjusting to font-size. I did some more testing, it looks like you are right and the height and with are only being honored by the Icon. and when applying it as a style. I'll go ahead and remove lines 82-86 and leave the accessory sizing to the consumer.

Then, I am thinking update the prop to be `accessoryWidth` that is applied. The ItemView's maxWidth keeps the width with the 60px maxWidth.

mjhenkes:
Could you add a width/height of 100% on the start accessory icon for both the all and all top aligned examples in site?

mjhenkes:
Could we remove the `reserveEndAccessorySpace` prop?

packages/terra-clinical-item-view/src/ItemView.scss
```diff
@@ -11,20 +11,44 @@
...
+
+  .start-accessory > * {
+    max-height: 2.8571rem;
+    max-width: 2.8571rem;
+  }
+
+  .end-accessory,
+  .reserve-end-accessory-space {
+    justify-content: flex-end;
+    max-height: 7.14285rem;
```
mjhenkes: It seems weird to have different max heights between start and end. Maybe bump this one back to 2.8571 rem?

mjhenkes: Do we have a max height for the item view itself?

emilyrohrbough: There is no maximum height for the item view. I can bump it back to 2.8571 rems. I wasn't sure what might be the best here so gave it 100px

mjhenkes: Could we pin the hight of the start and end accessories to the hight of the item view based off the text height?

emilyrohrbough:
@mjhenkes I could remove the `reserveEndAccessorySpace` I was using the prop to gain alignment with the Item Collection (see below), which was why I included it. I could see there being a concern about the end alignment being consistent, though maybe we move this spacing to the Item Collection Item to handle?

screen shot 2017-11-14 at 3 05 45 pm

mjhenkes:
Since the end accessory is dynamic, whatever end space we would reserve may be incorrect. The table will have to reserve the column though if there is an end accessory.

packages/terra-clinical-item-view/src/ItemView.scss
emilyrohrbough: When you say text height, are you thinking the height of the displays provided?

packages/terra-clinical-item-view/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* Added reserveStartAccessorySpace, and reserveEndAccessorySpace to allow for proper spacing when an accessory is not provided.
+
+### Changed
+* Updated the start accessory's max-width and mix-height to be 2.8571rem (40px)
```
kschuste: max-height instead of mix-height

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -48,24 +56,30 @@ const defaultProps = {
...
 };
 
 class ItemView extends React.Component {
 
-  static renderAccessory(accessory, accessoryAlignment) {
+  static renderAccessory(accessory, reserveSpace, accessoryAlignment, type) {
     let accessorySection;
-    if (accessory) {
+    if (accessory || reserveSpace) {
+      const reserveSpaceClassName = reserveSpace ? `reserve-${type}-accessory-space` : null;
```
kschuste: do we really want null to be in the classname?  wouldn't an empty string '' be better?

packages/terra-clinical-item-view/tests/nightwatch/item-view-spec.js
```diff
@@ -1,47 +1,58 @@
+/* eslint-disable no-unused-expressions */
 // eslint-disable-next-line import/no-extraneous-dependencies
 const resizeTo = require('terra-toolkit/lib/nightwatch/responsive-helpers').resizeTo;
 
 module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous'], {
   'Displays a clinical item view with default props': (browser) => {
     browser.url(`${browser.launchUrl}/#/tests/item-view-tests/default`);
-    browser.assert.elementPresent('#ItemView');
+    browser.expect.element('#ItemView').to.be.present;
```
kschuste: could remove the no-unused-expressions at the top for the linter and add a .to.equal(true) after each line of checking whether the element is present

browser.expect.element('#ItemView').to.be.present.to.equal(true)

packages/terra-clinical-item-view/src/ItemView.jsx
emilyrohrbough: nope we def don't

packages/terra-clinical-item-view/src/ItemView.scss
mjhenkes: yes + comment

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -48,24 +56,30 @@ const defaultProps = {
...
+
       const accessoryClassNames = cx(
         'accessory',
+        `${type}-accessory`,
+        `${reserveSpaceClassName}`,
```
jmsv6d: You should be able to just use `reserveSpaceClassName` without the template string.

packages/terra-clinical-item-view/src/ItemView.scss
```diff
@@ -11,20 +11,44 @@
     word-wrap: break-word;
   }
 
-  // Accessory is of fixed size and does not shrink or grow.
   .accessory {
+    display: flex !important;
```
jmsv6d: Why is the `!important` needed?

packages/terra-clinical-item-view/src/ItemView.scss
emilyrohrbough: So, I did some research. Ultimately, flexbox doesn't allow you t create a row of flexbox items with a max height defined by one child. It looks like there could be possible work arounds using position absolute & relative, but this isn't a flexbox solution. One suggestion I read was to switch from a flexbox layout and use a tabular layout instead. Not sure how complex that could get with this implementation having flexboxs within flexboxs.

emilyrohrbough: @JakeLaCombe Do you have any insights here? I know you're pretty familiar with flexbox.

packages/terra-clinical-item-view/src/ItemView.scss
emilyrohrbough: I check & its not. I removed it

packages/terra-clinical-item-view/tests/nightwatch/item-view-spec.js
jmsv6d: I prefer disabling the lint rule.

PaulSolDev:

emilyrohrbough:
@PaulSolDev pointed this out earlier for Edge, this was a reflection of this PR

screen shot 2017-11-17 at 1 18 17 pm

PaulSolDev:
Functional review complete.  The issue previously identified has been corrected.

PR #225, opened 2017-11-15
Add collapsible content prop.

dkasper-was-taken:
### Summary
Add collapsible content section to header and action header.

Questions:
What should the new prop be named?
What max-width should be set? Currently 50%.

Address: #192 

@cerner/terra


kschuste:
one item missing is the ability for the header to grow vertically rather than truncate the title text

dkasper-was-taken:
@kschuste Do we still want that functionality? I have to imagine that would be broken on mobile.

packages/terra-clinical-action-header/src/ActionHeader.jsx
```diff
@@ -59,6 +59,7 @@ const propTypes = {
 
   /**
    * Child element to be displayed on the right end of the header.
+   * This is intended to be used with the CollapsibleMenuView.
```
kschuste: your source code doesn't match the lib code, not sure if it matters if you don't recompile though but would be nice

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -8,6 +8,16 @@ const cx = classNames.bind(styles);
...
 const propTypes = {
   /**
+   * Child element to be displayed on the right end of the header.
+   */
+  children: PropTypes.element,
+  /**
+   * Content to be displayed at the end of the header.
+   * This can only be a collapsible react element.
+   */
+  collapsibleContent: PropTypes.element,
```
kschuste: don't see collapsibleContent being used below anywhere

dkasper-was-taken: Good catch.

kschuste:
can you add an example of this working with the header as well

packages/terra-clinical-action-header/src/ActionHeader.jsx
```diff
@@ -59,6 +59,7 @@ const propTypes = {
...
    */
   children: PropTypes.element,
```
kschuste: children implies that I can pass in an array of elements, this is allowing me to only pass in 1, should we allow this to be of type node, else this this name needs to be changed to child

dkasper-was-taken: Children is actually a keyword, so the name needs to remain the same. I don't mind changing it to node, but in the war room we discussed leaving the prop value at element until someone requests the functionality to pass additional elements.

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -8,6 +8,11 @@ const cx = classNames.bind(styles);
 
 const propTypes = {
   /**
+   * Child element to be displayed on the right end of the header.
```
jmsv6d: There should be something in the prop descriptions to differentiate between when children should be used and when startContent should be used.

PaulSolDev:
At the smallest allowable browser width for Edge I am seeing the right side of the Action Header cut off and no horizontal scroll bar available.
image

dkasper-was-taken:
@PaulSolDev That appears to be a limitation of the Edge browser, below 372px Edge no longer adjusts the deviceWidth associated with document body. It even loses its own collapsible button for it's toolbar at that breakpoint due to the minimum.

PaulSolDev:
@dkasper-was-taken Thank you for the explanation.  All other browsers checked out accordingly.  Functional review complete.

PR #226, opened 2017-11-15
Terra toolkit upgrade

ryanthemanuel:
### Summary
Upgrades toolkit to 2.2.0. This moves all testing into docker containers and enables webdriver.io testing.


emilyrohrbough:
Update Readme like so? https://github.com/cerner/terra-core/pull/1021/files#diff-04c6e90faac2675aa89e2176d2eec7d8R134

PR #247, opened 2018-01-09
Item Display: New styles

### LINKED ISSUE #209: Item Display: add new styles
# Issue Description
1. Need to add two new textStyles to Item Display:
   1a. that makes text bold but keeps current text color (not red).
   1b. an inactive version, superlight text 
2. Change the Item Comment icon color to use transparency (looks better when row is selected).

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Add two new options to the textStyle prop: 'bold'/'strong', and 'inactive'/'disabled' or similar wording for the props.

## Current Behavior
Item Display currently has {default}, 'secondary', 'attention', 'strikeThrough'.

noahbenham:
### Summary
Resolves #209. Adds new text style and `isInvalid` prop to Item Display and improves examples.

### Deployment Link
https://terra-clinical-deployed-pr-247.herokuapp.com/static/#/site/item-display


packages/terra-clinical-item-display/src/ItemDisplay.scss
```diff
@@ -39,6 +39,14 @@
     font-weight: 700 !important;
```
noahbenham: Team: Why do these have `!important`? This seems potentially restrictive to consumers.

jmsv6d: I don't see any reason why it would be needed. My vote would be to remove it.

neilpfeiffer:
Added a comment to issue that will effect tech design - suggestion is to change "inactive" from being an additional `textStyle` to a new prop `isInvalid`:   https://github.com/cerner/terra-clinical/issues/209#issuecomment-356362914

Also added the new colors.

mjhenkes:
@neilpfeiffer, Is the color contrast of item display with isInvalid enabled accessible?

packages/terra-clinical-item-display/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Add `strong` & `inactive` text styles.
```
emilyrohrbough: This needs to be updated. It should listed as `Added` since they didn't previously exist and we want to indicate `isInvalid` is a new prop.

packages/terra-clinical-item-display/docs/README.md
```diff
@@ -18,13 +18,13 @@ import ItemDisplay from 'terra-clinical-item-display';
   text="display text"
   textStyle="attention"
   isTruncated
-  icon=<img alt="Graphic" />
+  isInvalid
```
emilyrohrbough: Lets update these examples to reflect a 'truer' implementation where `isInvalid={this.state.isValid}`

packages/terra-clinical-item-display/src/ItemDisplay.jsx
```diff
@@ -19,13 +20,17 @@ const propTypes = {
    */
   text: PropTypes.string,
   /**
-   * The visual styles to be applied to the display element.
+   * The visual style to be applied to the display element. One of: `secondary`, `attention`, `strikeThrough`, `strong`.
```
emilyrohrbough: nit: add or between strikeThrough and strong

packages/terra-clinical-item-display/src/ItemDisplay.jsx
```diff
@@ -36,25 +41,29 @@ const defaultProps = {
...
   ]);
 
   let displayIcon;
   if (icon) {
-    displayIcon = <div className={cx('inline-icon')}>{icon}</div>;
+    const iconClasses = cx('inline-icon', { 'is-invalid': isInvalid });
+    displayIcon = <div className={iconClasses}>{icon}</div>;
   }
 
   return (
```
emilyrohrbough: I feel like we should place the 'is-invalid' class on the Item-Display div instead of the text and icon divs. Is there a reason we apply it twice?

emilyrohrbough:
There is [this issue](https://github.com/cerner/terra-core/issues/1044) logged in terra-core to export prop options through the Utils object. Would you mind making that change here for the `textStyles` prop?

emilyrohrbough:
I took a look at the deployment link. I am uncertain that the the `isInvalid` styling will pass accessibility color contrast. @neilpfeiffer 

We should get some webdriver.io accessibility tests for this component. Here is [an example](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-error-view/tests/wdio/error-view-spec.js) of what those tests would look like. Visual Regression testing is still an in-progress piece so I wouldn't worry about writing snapshot test steps, but accessibility test steps are straight forward.

packages/terra-clinical-item-display/src/ItemDisplay.jsx
noahbenham: Yeah I considered that and went with this approach since it was already being applied directly to the text instead of surrounding div. I agree with you though. I'll update this.

packages/terra-clinical-item-display/docs/README.md
noahbenham: Just clarifying.. Are you suggesting using `this.state.isValid` which would be defined by a consumer in our example? The current example indicates this is a bool with no additional cognitive load of determining what state means.

noahbenham:
I added a `wdio` test which failed due to [color-contrast](https://dequeuniversity.com/rules/axe/2.5/color-contrast?application=axeAPI). Adjusting opacity from `.18` to `.2` fixed the issue, which should be an okay value per discussions with @neilpfeiffer a couple days ago. See 76917949fcb9612445cf00f791b56ba20b6c25c3.

packages/terra-clinical-item-display/tests/wdio/item-display-spec.js
```diff
@@ -0,0 +1,10 @@
+/* global browser, describe, it, expect, before, Terra */
+const viewports = Terra.viewports('tiny', 'medium');
+
+describe('ItemDisplay', () => {
+  describe('disabled', () => {
```
emilyrohrbough: Can we add a wdio-test for each route? so 5 more?

neilpfeiffer:
Late response to @mjhenkes, @emilyrohrbough, @noahbenham: 

As a general rule - inactive or disabled items within a UI do not need to pass any accessibility guidelines: https://ux.stackexchange.com/a/80885

_(Really not even sure why changing the opacity from 0.18 to 0.2 passes, unless when color sampled there is an absolute most-bottom/minimum "1.5 : 1" contrast that is in the checker. #1C1F21 @ 0.2 certainly doesn't pass the 4.5:1 or 3:1 as a solid color on white.)_  

I'm fine with keeping it at 0.2 opacity for now and seeing how it looks in context of some example clinical item-views. Though I'm curious if the secondary textStyle color #75767D will also fail, even at 0.2, which is fine since they actually don't need to pass anyway.

noahbenham:
Deferring @emilyrohrbough's [comment](https://github.com/cerner/terra-clinical/pull/247#issuecomment-356683131) to [issue discussion here](https://github.com/cerner/terra-core/issues/1044) as implementing it created circular dependency issues and larger questions for me about how we want to proceed with that change.

I've created the additional `wdio` tests requested. Neil's comment indicates they're unnecessary/optional. Keeping them per offline discussions with @emilyrohrbough. I don't think `wdio` tests for `/tests/item-display-tests` has value, so skipping that one.

packages/terra-clinical-item-display/src/ItemDisplay.jsx
```diff
@@ -19,13 +20,17 @@ const propTypes = {
...
-   * The visual styles to be applied to the display element.
+   * The visual style to be applied to the display element. One of: `secondary`, `attention`, `strikeThrough`, or `strong`.
    */
   textStyle: PropTypes.oneOf(textStyles),
   /**
    * Whether or not the text should be truncated.
    */
   isTruncated: PropTypes.bool,
+  /**
+   * Whether or not the text should be styled as disabled.
```
emilyrohrbough: nit: display is disabled

packages/terra-clinical-item-display/tests/jest/ItemDisplay.test.jsx
```diff
@@ -24,11 +24,22 @@ it('should render with text and graphic', () => {
   expect(itemDisplay).toMatchSnapshot();
 });
 
+it('should render disabled with text and graphic', () => {
```
emilyrohrbough: nit: ...disabled __display__ with...

noahbenham:
Item comment icons now have `color: rgba(117, 118, 125, 0.3)` per Neil's request. Let me know if anything needs additional tests.

neilpfeiffer:
Tested: PR #247
Resolves: Issue #209

**Reviewed:**
- Follows UX enhancements specified in #209.

\
Verified by @neilpfeiffer, Ready for Release after functional review.


**Pass** |
-----------|
UX Review|

PaulSolDev:
Functional review complete. No issues identified.

Issue #252, opened 2018-01-17
Deprecate Error and No Data Views


kschuste:
# Issue Description
Deprecate terra-error-view and terra-no-data-view.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [X] Other

## Expected Behavior
terra-error-view and terra-no-data-view are removed.  Have been replaced by the terra-status-view in the terra-core project.

noahbenham:
Thanks for opening this, I was just about to consume Error View and will hold off in favor of Status View. When will `terra-status-view` be available on npm?

ChaseSinclair:
@kschuste We are currently using the Error View for the new HealtheatCerner site [here](https://github.cerner.com/HealtheLife/healthe-at-cerner/blob/master/package.json#L54). Can we keep this from being deprecated?

dm123455:
I am also curious why the search did not identify our repo!

kschuste:
@ChaseSinclair you will still be able to use the error view, but it will still be deprecated, I would advise to switch to the terra-status-view in the terra-core repo when is released.

mjhenkes:
oops didn't mean to close that

PR #253, opened 2018-01-17
deprecate error-view and no-data view from repository

kschuste:
### Summary
This resolves issue #252.
deprecating terra-error-view and terra-no-data-view.

### Additional Details
terra-error-view and terra-no-data-view have been replaced by the terra-status-view in the terra-core project.  Therefore, can deprecate these components.


mjhenkes:
I know I told you to do this, but I think i changed my mind. Can we set the packages to hidden and update the readme's for terra clinical site to say that the components are depreciated and point to terra-status? 
Heres an example of how we set terra-button 2.0 to hidden. https://github.com/cerner/terra-core/blob/master/lerna.json

mjhenkes:
Is there an easy way to also disable running tests on these packages?

kschuste:
@mjhenkes good idea, we are able to exclude the paths in the configuration setups to keep them from running

packages/terra-clinical-error-view/README.md
```diff
@@ -4,7 +4,8 @@
 [![NPM version](http://img.shields.io/npm/v/terra-clinical-error-view.svg)](https://www.npmjs.org/package/terra-clinical-error-view)
 [![Build Status](https://travis-ci.org/cerner/terra-clinical.svg?branch=master)](https://travis-ci.org/cerner/terra-clinical)
 
-The Terra ErrorView is a display presented with an error message and location for a button.
+The Terra ErrorView is a display presented with an error message and location for a button.  This component has been deprecated in
```
mjhenkes: Could we bold the 'deprecated' sentence for all these readmes?

kschuste: deprecated is now in bold

PR #265, opened 2018-01-31
Header spacing between buttons & text

### LINKED ISSUE #264: Header start & end margins
# Issue Description

Would it make sense to include default margins on either side of the left & right header content, or is that too restrictive to consumers?

I'd like to be able to just drop in text and buttons on either side but with the current design, it looks like this:

screen shot 2018-01-31 at 12 25 45 pm

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior

Header adds margin between left/right content & text.

## Current Behavior

No margin is specified - even Terra examples do this manually to make things look okay.

## Environment

* Component Version: 1.7.0
* Browser Name and Version: N/A
* Operating System and version (desktop or mobile): N/A

noahbenham:
### Summary
Resolves #264. This adds additional spacing between Header buttons and text.

### Additional Details
This gets a Heroku demo setup for design review. Additional `wdio` or snapshot tests may be needed. Changes to margin amount or which element margins are applied to may be needed.

### Deployment Link
https://terra-clinical-deployed-pr-265.herokuapp.com/static/#/site/header


emilyrohrbough:
@neilpfeiffer @StephenEsser and I were discussing offline. Adding this change would create weird visuals for consumers already adding the appropriate spacing to handle this. Just something we should probably consider.

noahbenham:
@emilyrohrbough What type of tests would you recommend to ensure this spacing doesn't regress?

noahbenham:
@jmsv6d I've adjusted this to apply styling only to the left or right content if it exists. It looks good locally but not deployed; I'm not sure why.

Action Header uses `1.429rem` whereas this uses `5px`. We probably need design input from @neilpfeiffer for consistent sizing between Header & Action Header.

JakeLaCombe:
Should we use Terra Space for this? https://terra-core-deployed.herokuapp.com/static/#/site/spacer

mjhenkes:
@JakeLaCombe we discussed using the spacer here https://github.com/cerner/terra-clinical/issues/264. 

Quoting myself 😉 
>The spacer works, but this is a primary use case for this component, and by requiring use of the spacer component the spacing for the header won't be theme-able except at a broad level.

JakeLaCombe:
Thanks for the clarification!

packages/terra-clinical-header/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
```
emilyrohrbough: This should be added

packages/terra-clinical-header/src/Header.scss
```diff
@@ -39,6 +39,14 @@
     position: relative;
   }
 
+  .start-content {
+    margin-right: 5px;
```
emilyrohrbough: These should be updated to be rem values

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -45,7 +45,7 @@ const defaultProps = {
 const Header = ({ children, title, startContent, endContent, isSubheader, ...customProps }) => {
   let startElement;
   if (startContent) {
-    startElement = <div className={cx('flex-end')}>{startContent}</div>;
+    startElement = <div className={cx('start-content flex-end')}>{startContent}</div>;
```
emilyrohrbough: the styling for this isn't being applied. To fix this, update how you are binding these class names.

Should be:
`className={cx(['start-content', 'flex-end'])}

packages/terra-clinical-header/src/Header.scss
noahbenham: This needs design input; I'll update it as soon as I have a rem value from UX.

emilyrohrbough:
@noahbenham As @jmsv6d mentioned, this is going to have a cascading affect on the terra-clinical-action-header component. We are going to need updated padding values for ActionHeader to handle this PR change. Additionally,  after looking at the action header styling, that component will need some updates to handle single vs multiple start and end content.

emilyrohrbough:
@noahbenham It looks like the Header test link are not included in the site. Could you add those to this [file](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-site/src/TestLinks.jsx#L6)?

.gitignore
```diff
@@ -57,3 +57,4 @@ target
 **/__snapshots__/screen
 **/__snapshots__/diff
 errorScreenshots
+.DS_Store
```
ryanthemanuel: 👍

ryanthemanuel: We should add this to the generator(s)

mjhenkes:
We should close this pr and address this when the action header/ header is moved to core

noahbenham:
@mjhenkes If we close this can we release that component with this spacing baked in from the start to prevent affecting consumers on a minor release?

Also any idea of an ETA for that? Just curious.

mjhenkes:
@noahbenham, the spacing should be built in from the start. I had heard that @alex-bezek was looking at taking on some of the work to move the header to core. I will need to sync up with him.

PR #272, opened 2018-02-12
React 16 uplift

### LINKED ISSUE #161: React 16 upgrade
# Issue Description
When react 16 comes out, we'll want to enable compatibility with react 16.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Terra components are compatible with React 16

## Current Behavior
N/A

benbcai:
### Summary
Closes #161 

### Additional Details
- Changes are similar to https://github.com/cerner/terra-core/pull/1194
- Upgraded packages to react and react-dom 16.2.0.
- Use enzyme-adapter-react-16
- Added raf-polyfill for requestAnimationFrame.
- Updated wdio.conf.js to be more in line with terra-core.
- Added a test container with a selector for webdrive tests to capture screenshots since [data-reactroot] is removed.
- Reference WDIO screenshots terra-clinical-item-display are regenerated because it used to capture the entire screen instead of just the selector.
- terra-clinical-error-view and terra-clinical-no-data-view have been deprecated so they are left as is.
- Bumped all the terra-core components to their new major version.

### Deployment Link
https://terra-clinical-deployed-pr-272.herokuapp.com/static/#/site


bjankord:
@benbcai You'll want to bump all the terra-core components to a new major version in the package.json files. Most should be bumped to ^2.0.0, but some may need to be bumped higher. I believe base needs to be bumped to ^3.0.0 and terra-grid needs to be bumped to ^4.0.0

benbcai:
@bjankord Now that the 2.0 versions just got released, I can update to pull those in.

packages/terra-clinical-application-site/package.json
```diff
@@ -27,10 +27,9 @@
   },
   "dependencies": {
     "prop-types": "^15.5.8",
-    "react-intl": "^2.2.3",
-    "react-modal": "^1.6.5",
```
benbcai: I think it is safe to remove `react-modal` since we don't use it anymore in the demo app. We now use terra-modal-manager.

jestconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "setupFiles": [
+  	"raf/polyfill",
```
emilyrohrbough: nit: extra space

wdio.conf.js
```diff
@@ -43,10 +46,21 @@ const config = {
...
     },
   },
+  terra: {
+    selector: '#wdio-test-wrapper',
+  },
 
    // Configuration for WebPackDevService
   webpackPort,
   webpackConfig,
+
```
emilyrohrbough: Lets not add this global refresh() hook. We don't need it at this point. Plus this is something we will want to remove from our webdriver.io configuration. If a test step needs to refresh after loading the browser, it should be specified within that test step that needs it.

wdio.conf.js
```diff
@@ -23,7 +23,10 @@ const config = {
   ],
 
   // Terra-toolkit services are 'visual-regression', AxeService, TerraService, SeleniumDockerService
-  services: wdioConf.config.services.concat([WebpackDevService]),
+  // services: wdioConf.config.services.concat([WebpackDevService]),
```
emilyrohrbough: Oops, when you move this to the bottom, you didn't remove this comment.

wdio.conf.js
benbcai: Removed: https://github.com/cerner/terra-clinical/pull/272/commits/3e968067da3a39875c1b528dd8615f112b61fd72

wdio.conf.js
benbcai: Removed: https://github.com/cerner/terra-clinical/pull/272/commits/3e968067da3a39875c1b528dd8615f112b61fd72

benbcai:
@bjankord The terra-core components are bumped. Do you think it is better to just manually bump all the components in terra-clinical in this PR or do you prefer to let lerna bump them during publish? If we bump during publish we would run into the issue where peer dependencies won't get bumped and we would still have to bump the peer dependencies manually.

bjankord:
@benbcai We should be able to try a dry-run with lerna and see if it bumps the peerDependencies or not. It seems to bump them for minor releases, not sure if there is a bug in lerna with bumping peerDeps when doing a major release.

mjhenkes:
Make sure you update the readme to mark terra-clinical-application as deprecated

benbcai:
I only removed the demo app and not terra-clinical-application. terra-clinical-application is now also removed and is marked as deprecated on the [readme](https://github.com/cerner/terra-clinical/tree/react-16-uplift#status)

packages/terra-clinical-action-header/package.json
```diff
@@ -22,26 +22,26 @@
...
+    "terra-icon": "^2.0.0"
   },
   "dependencies": {
     "classnames": "^2.2.5",
     "prop-types": "^15.5.8",
-    "terra-base": "^2.0.0",
-    "terra-button": "^1.0.0",
-    "terra-button-group": "^1.0.0",
+    "terra-base": "^3.0.0",
+    "terra-button": "^2.0.0",
```
JakeLaCombe: We should bump these to their minor versions to ensrue peer dependencies work ok. https://github.com/cerner/terra-core/commit/71f8b9870f142fb23bf757611e689b6038062b1e

benbcai: Minor versions are bumped: https://github.com/cerner/terra-clinical/pull/272/commits/f0dd30ac2b6797a81fe54590e16f6555bd5b561f

benbcai:
@PaulSolDev This is ready for functional review. I'll sync up with you on what all needs to be reviewed.

PaulSolDev:

PR #273, opened 2018-02-12
Add option to display close button on small viewports

yusufali2205:
### Summary
This gives option to the consumers to opt out of the default close button behavior on small view ports in terra-clinical-action-header. https://github.com/cerner/terra-clinical/blob/terra-clinical-header%401.7.0/packages/terra-clinical-action-header/src/ActionHeader.jsx#L27

### Additional Details
The default behavior is:
"On small viewports a back button will be displayed instead of a close button when a separate onBack callback is not set"

This change adds a new boolean prop `keepCloseButton` (you can suggest a better name) which when set true will disable the default behavior.

### DIFF
--- a/packages/terra-clinical-action-header/CHANGELOG.md
+++ b/packages/terra-clinical-action-header/CHANGELOG.md
@@ -4,7 +4,7 @@ ChangeLog
 Unreleased
 ----------
 ### Added
-* prop to keep the close button in small viewports
+* prop `keepCloseButton` to keep the close button in small viewports
 
 1.6.0 - (December 9, 2017)
 ----------
--- a/packages/terra-clinical-action-header/CHANGELOG.md
+++ b/packages/terra-clinical-action-header/CHANGELOG.md
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* prop to keep the close button in small viewports
 
 1.6.0 - (December 9, 2017)
 ----------
--- a/packages/terra-clinical-action-header/src/ActionHeader.jsx
+++ b/packages/terra-clinical-action-header/src/ActionHeader.jsx
@@ -58,6 +58,11 @@ const propTypes = {
   onPrevious: PropTypes.func,
 
   /**
+   * A Boolean indicating if close button should be displayed on small viewports when separate onBack callback is not set.
+   */
+  keepCloseButton: PropTypes.bool,
+
+  /**
    * Child element to be displayed on the right end of the header.
    * This is intended to be used with the CollapsibleMenuView.
    */
@@ -72,6 +77,7 @@ const defaultProps = {
   onMinimize: null,
   onNext: null,
   onPrevious: null,
+  keepCloseButton: null,
   children: null,
 };
 
@@ -92,6 +98,7 @@ const ActionHeader = ({
   onMinimize,
   onPrevious,
   onNext,
+  keepCloseButton,
   children,
   ...customProps }, {
   intl,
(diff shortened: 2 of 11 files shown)


packages/terra-clinical-action-header/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* prop to keep the close button in small viewports
```
StephenEsser: Please specify the prop name that was added

bjankord:
> This is a fairly significant change from a requirements and workflow perspective. Has this change been discussed / approved by our UX and solution designers?

Yeah we need to check on this. I found out the other day that this is blocking us on using action header in the dialog. We need a close button in the modal in narrow screens. What was logic on switching to a back button and hiding the close button in the action header originally?

yusufali2205:
@StephenEsser I have not discussed this change with UX / SD. We are using the action-header but the back button in our case looks confusing. We needed an option to opt out of this behavior, otherwise we will have to switch to terra-clinical-header.

packages/terra-clinical-action-header/CHANGELOG.md
yusufali2205: done 6eae453

mjhenkes:
I agree with this change, I think it's a bit heavy handed to for the close back arrow functionality.

packages/terra-clinical-action-header/src/ActionHeader.jsx
```diff
@@ -72,6 +77,7 @@ const defaultProps = {
   onMinimize: null,
   onNext: null,
   onPrevious: null,
+  keepCloseButton: null,
```
emilyrohrbough: This should default to false.

yusufali2205: done b36d87c

packages/terra-clinical-site/src/examples/action-header/Index.jsx
```diff
@@ -24,6 +25,8 @@ const HeaderExamples = () => (
     <ActionHeaderBackNextPreviousClose />
     <h2>Maximize, Close</h2>
     <ActionHeaderExpandClose />
+    <h2>Close, keepCloseButton</h2>
```
benbcai: Non-technical audience may not understand what this example does. Perhaps rephrase keepCloseButton to something that is a bit more descriptive.

yusufali2205: you mean I should change the [property name](https://github.com/cerner/terra-clinical/pull/273/files#diff-bb6aa0f3cc7f39549fc68570eab3f3c4R63) or just add explanation in this example?

benbcai: Just the description in the example. Instead of "Close, keepCloseButton", perhaps "Keeps Close button in small view port" or something similar.

packages/terra-clinical-action-header/src/ActionHeader.jsx
```diff
@@ -109,7 +116,7 @@ const ActionHeader = ({
 
   let closeButtonSmall;
   let backButtonSmall;
-  if (onClose && !onBack) {
+  if (onClose && !onBack && !keepCloseButton) {
```
jmsv6d: This approach works, but when this prop is set it removes the need for the responsive element. So a better solution would be to just render the actionHeader without the responsive element when this prop is set, so that we don't have that resize observer when we don't need it.

jmsv6d: Actually, nevermind. It's also needed for the expand/collapse logic.

jmsv6d:
Tests should be added for this.

PaulSolDev:

mjhenkes:
This is available on terra-clinical-action-header@1.7.0

PR #285, opened 2018-03-01
Migrate all but detail view and clinical item display

mjhenkes:
### Summary
Except detail view and clinical item display until https://github.com/cerner/terra-dev-site/issues/9 is addressed.

@cerner/terra


mjhenkes:
Also this is 100% validiated by passing tests, since i just moved file locations arounds

PR #286, opened 2018-03-01
Add wdio tests for terra-clinical-detail-view

rakshith91:
Resolves #286 

@cerner/terra


packages/terra-clinical-detail-view/tests/wdio/DetailList/detail-list-spec.js
```diff
@@ -0,0 +1,15 @@
+/* global browser, Terra, before */
+
+const viewports = Terra.viewports('tiny', 'small', 'medium', 'large', 'huge');
+const selector = '#root';
+
+describe('Detail List', () => {
+  describe('Displays a default Detail List with a title and list', () => {
```
emilyrohrbough: In the old nightwatch, there wasn't a test for the `large-item-list` example as it was proving the list flex for large display items. Could we add a test step to capture this behavior?

rakshith91: c981c852fc2c0da99711b31915c67acce79fb78c

PR #316, opened 2018-03-29
Update lower bound express version 

emilyrohrbough:
### Summary
Sync with https://github.com/cerner/terra-core/issues/1375


mjhenkes:
We should supply an express server with either terra-toolkit or terra-dev-site.

emilyrohrbough:
@mjhenkes I am working on updating the dev-site's express script to handle the use from other repos since we already ship its scripts with the package

PR #318, opened 2018-03-30
[DO NOT MERGE] Learning Docker & Curious

emilyrohrbough:
### Summary
I'm just testing a few things I'm curious about

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ branches:
   only:
     - master
 cache: false
-before_install:
-  - docker pull selenium/standalone-chrome:3.8.1-erbium
-  - docker run -v /dev/shm:/dev/shm -e TZ=America/Chicago -d -p 4444:4444 selenium/standalone-chrome:3.8.1-erbium
 before_deploy:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,9 +7,6 @@ branches:
   only:
     - master
 cache: false
-before_install:
```
emilyrohrbough: We do not need to specify this in our travis configuration because this is specified in the toolkit's [docker-compose](https://github.com/cerner/terra-toolkit/blob/master/src/wdio/docker-compose.yml). So we can actually utilize the same docker build locally and in our Travis CI builds, which reduces the places to specify our build setup.

wdio.conf.js
```diff
@@ -24,8 +24,6 @@ const config = {
 
   // Configuration for terra-toolkit's SeleniumDocker service
   seleniumDocker: {
-    enabled: !process.env.TRAVIS,
-    cleanup: false,
```
emilyrohrbough: Cleanup is not a valid seleniumDocker configuration key. This value is never considered.

PR #319, opened 2018-04-03
Clinical Result View

### LINKED ISSUE #267: Add Clinical Result View
# Issue Description
Add reusable view component currently called 'ClinicalResultView' that is used to show measurements such as labs and vitals in various views (grids/lists/stand alone).

example crv

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

rakshith91:
### Summary
Resolves #267 
The Clinical Result View is a small stateless view that is intended to show measurements such as labs and vitals. The view is designed in a way that it can handle a single measurement (such as weight), but also show paired measurements such as blood pressure. In a clinical setting, each measurement can have additional metadata associated on the result, like whether it's a critical value or has been modified, or even if it's an unexpected value.

### Sample Images
image
_________________
image

@cerner/terra


packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,195 @@
...
+
+const resultIndicator = (result) => {
+  let icon = null;
+
+  if (!result.normalcy || !result.normalcy.length) {
+    return null;
+  }
+
+  if (result.normalcy === 'ABNORMAL') {
+    icon = <IconAbnormal />;
```
yuderekyu: We can define/use constants for the normalcy, then export them like [this](https://github.com/cerner/terra-core/blob/master/packages/terra-avatar/src/Avatar.jsx#L101) for consumers to use.

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,195 @@
...
+  if (!results || !results.length) {
+    const noValueDisplay = '--';
+    return <div className={cx('result-view-result-value-text')} >{noValueDisplay}</div>;
+  }
+
+  const resultsDivs = [];
+
+  for (let i = 0; i < results.length; i += 1) {
+    if (resultsDivs.length >= 1) {
+      const seperator = <span className={cx('result-view-result-separator')} >{'/'}</span>;
```
yuderekyu: Nit: *separator

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -0,0 +1,114 @@
+@import './mixins';
+
+:local {
+  .result-view {
+    color: var(--result-view-background-color, #595f62);
+    font-size: 20px;
```
yuderekyu: Do we want font-size and line-height themeable here and in the other css classes?

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -0,0 +1,114 @@
...
+      fill: currentColor;
+    }
+
+    &.align-end {
+      text-align: right;
+    }
+  }
+
+  .result-view-body--padded {
+    padding: 5px;
```
yuderekyu: Same themeable question for the padding here and in subsequent classes.

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,195 @@
...
+import IconLow from 'terra-icon/lib/icon/IconLow';
+import styles from './ResultView.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   *  An array of one or more results.
+   */
+  results: PropTypes.array,
```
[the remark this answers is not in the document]
rakshith91: I think results can accept other attributes too. I'll sync up with @jeffmerten before making this change.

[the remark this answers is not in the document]
rakshith91: Changed it.

packages/terra-clinical-result-view/tests/wdio/result-view-spec.js
```diff
@@ -0,0 +1,46 @@
+/* global browser, Terra */
```
yuderekyu: We should also be testing if the themeable variables are applied.

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,189 @@
...
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   *  An array of one or more results.
+   */
+  results: PropTypes.arrayOf(PropTypes.shape({
+    value: PropTypes.number,
+    units: PropTypes.string,
```
yuderekyu: Oops, this should be *unit.

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,189 @@
...
+  hasDocument: false,
+  hasComment: false,
+  isModified: false,
+  alignEnd: false,
+  isTruncated: true,
+  isPadded: false,
+};
+
+const normalcyIndicators = {
+  ABNORMAL: <IconAbnormal />,
```
yuderekyu: On second thought, do we know if the normalcy indicators are constrained to these four icons, or should we be agnostic on what the indicators are? If that's the case, we could change the `normalcy` prop to a type `node`, and let the consumer pass in whatever icon they want. Thoughts?

jeffmerten: We should be agnostic on the indicators and rgb values in the terra-clinical-result-view. From my investigation I found the FHIR spec for observation has quite a large set of what can be considered normalcy. https://www.hl7.org/fhir/valueset-observation-interpretation.html

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -0,0 +1,114 @@
...
+
+  .result-view-result-block {
+    align-items: baseline;
+    display: inline-flex;
+  }
+
+  .result-view-indicator {
+    padding-right: var(--result-view-indicator-padding-right, 3px);
+
+    &.high {
```
jeffmerten: The normalcy colors should be provided by the consumer.

packages/terra-clinical-result-view/src/ResultView.jsx
rakshith91: copy pasting is too powerful for me :(

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,173 @@
...
+      resultsDivs.push(separator);
+    }
+
+    const resultValue = results[i];
+    const resultIndicatorDiv = resultIndicator(resultValue);
+    if (resultIndicatorDiv) {
+      resultsDivs.push(resultIndicatorDiv);
+    }
+
+    if (resultValue.unexpected) {
```
jeffmerten: I don't see anywhere in the proptypes where this is declared.

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,173 @@
...
+   */
+  isModified: PropTypes.bool,
+  /**
+   *  Are there comments related to the result. Adds an indicator.
+   */
+  hasComment: PropTypes.bool,
+  /**
+   *  Is the view aligned to the right. Defaults to false.
+   */
+  alignEnd: PropTypes.bool,
```
jeffmerten: Wondering if this should be called alignment, with two options. This would make it easier to support a center option in the future if necessary.

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,173 @@
...
+
+const propTypes = {
+  /**
+   *  An array of one or more results.
+   */
+  results: PropTypes.arrayOf(PropTypes.shape({
+    value: PropTypes.number,
+    unit: PropTypes.string,
+    normalcy: PropTypes.node,
+    normalcyColor: PropTypes.string,
```
jeffmerten: Instead of normalcy and normalcyColor, how about icon and color?

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,173 @@
...
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   *  An array of one or more results.
+   */
+  results: PropTypes.arrayOf(PropTypes.shape({
+    value: PropTypes.number,
+    unit: PropTypes.string,
+    normalcy: PropTypes.node,
```
jeffmerten: Can you do typeOf Icon rather than node?

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,173 @@
...
+    if (resultValue.unit && resultValue.unit.length > 0) {
+      unitDiv = <small className={cx('result-view-result-unit')} >{resultValue.unit}</small>;
+    }
+
+    let resultValueDisplay = '--';
+    if (resultValue.value) {
+      resultValueDisplay = resultValue.value;
+    }
+
+    let resultValueTextClassName = 'result-view-result-value-text';
```
jeffmerten: Could probably make this a ternary.

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,173 @@
...
+    if (resultValue.unexpected) {
+      resultsDivs.push(<IconUnexpected />);
+    }
+
+    let unitDiv = null;
+    if (resultValue.unit && resultValue.unit.length > 0) {
+      unitDiv = <small className={cx('result-view-result-unit')} >{resultValue.unit}</small>;
+    }
+
+    let resultValueDisplay = '--';
```
jeffmerten: Could probably simplify this to result.value || '--'

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -0,0 +1,62 @@
...
+
+  .result-view-result-age {
+    color: var(--result-view-text-color, #6f7477);
+    display: block;
+    font-size: var(--result-view-font-size, 20px);
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  .result-view-is-truncated {
```
jeffmerten: Nit: I'd get rid of "is"

packages/terra-clinical-result-view/src/ResultView.jsx
rakshith91: enforcing an svg icon

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,164 @@
...
+   */
+  isModified: PropTypes.bool,
+  /**
+   *  Are there comments related to the result. Adds an indicator.
+   */
+  hasComment: PropTypes.bool,
+  /**
+   *  The alignment of the component. Defaults to left.
+   */
+  alignment: PropTypes.oneOf(['right', 'center']),
```
jeffmerten: did you mean left instead of center?

rakshith91: Nope, I mean center. We don't need to specify left explicitly since it's the default alignment. I think it'll be a good idea to support both center and right.

packages/terra-clinical-result-view/tests/jest/ResultView.test.jsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+import ResultView from '../../src/ResultView';
+
+const locale = 'en-US';
+const results = [{ value: 503456789, unit: 'ml', normalcy: 'CRITICAL' }];
```
jeffmerten: looks like this is outdated

rakshith91: oops. my bad. fixed it.

packages/terra-clinical-result-view/src/ResultView.jsx
jeffmerten: I don't see that pattern being applied in other places. If a consumer wants to pass in 'left' explicitly, they should be able to.

https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-item-view/src/ItemView.jsx#L26
https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-item-view/src/ItemView.jsx#L57

rakshith91: Changed it

packages/terra-clinical-result-view/tests/wdio/result-view-spec.js
rakshith91: Added tests ( and addresses other comments)

packages/terra-clinical-result-view/src/_mixins.scss
```diff
@@ -0,0 +1,8 @@
+@mixin terra-clinical-text-truncate {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  width: var(--result-view-truncation-width, 30px);
```
yuderekyu: Let's also use `rem` for `width`.

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -0,0 +1,70 @@
...
+
+  .result-view-result-value-text {
+    font-size: var(--result-view-result-value-font-size, 20px);
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  .result-view-result-unit {
+    color: var(--result-view-result-unit-text-color, #6f7477);
+    line-height: normal;
```
yuderekyu: `line-height` defaults to `normal`, so we can remove this.

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -0,0 +1,70 @@
...
+      text-align: center;
+    }
+  }
+
+  .result-view-body--padded {
+    padding: var(--result-view-body-padding, 5px);
+  }
+
+  .result-view-result-block {
+    align-items: baseline;
```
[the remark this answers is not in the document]
rakshith91: The [TRVM implementation](https://github.cerner.com/ion-chart/orion-clinical_results/blob/master/app/assets/stylesheets/orion/clinical_results/clinical_result_view.less#L65) of result view uses `baseline`.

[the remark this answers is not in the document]
yuderekyu: Is that alignment what we want though? We can defer to UX, but here's what it looks like toggling `baseline` vs `center`:
ezgif-2-2eff3376bf

packages/terra-clinical-result-view/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "terra-clinical-result-view",
+  "main": "lib/ResultView.js",
+  "version": "0.0.0",
+  "description": "{insert description}",
```
StephenEsser: Missing a discription

packages/terra-clinical-result-view/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-clinical/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-clinical#readme",
+  "peerDependencies": {
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0",
+    "terra-base": "^3.1.0"
```
StephenEsser: I think you can remove this peer dependency if there is a hard dependency specified below

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,164 @@
...
+  if (!result.icon) {
+    return null;
+  }
+  icon = React.cloneElement(
+    result.icon,
+    { color: result.color },
+  );
+
+  return <span className={cx('result-view-indicator')} >{icon}</span>;
+};
```
StephenEsser: There are a couple of ways this could be reduced, here are a few possible examples: 

```
const resultIndicator = ({ icon, color }) => {
  if (icon) {
    const icon = React.cloneElement(icon, { color });
    return <span className={cx('result-view-indicator')}>{icon}/span>;
  }
  return null;
};
```

```
const resultIndicator = ({ icon, color }) => {
  if (!icon) {
    return null;
  }

  const icon = React.cloneElement(icon, { color });
  return <span className={cx('result-view-indicator')}>{icon}/span>;
};
```

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,164 @@
...
+
+const resultsDiv = (results, isTruncated) => {
+  if (!results || !results.length) {
+    const noValueDisplay = '--';
+    return <div className={cx('result-view-result-value-text')} >{noValueDisplay}</div>;
+  }
```
StephenEsser: Creating the variable makes this longer, what are your thoughts on changing this to:

```
if (!results || !results.length) {
  return <div className={cx('result-view-result-value-text')}>{'--'}</div>;
}
```

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,164 @@
...
+  const attributesName = 'result-view';
+  let align = 'alignment-left';
+
+  if (alignment === 'right') {
+    align = 'alignment-right';
+  } else if (alignment === 'center') {
+    align = 'alignment-center';
+  }
+
+  let bodyClassName = 'result-view-body';
```
StephenEsser: Is this className used anywhere? I don't see any CSS

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,164 @@
...
+
+  if (isPadded) {
+    bodyClassName = 'result-view-body--padded';
+  }
+
+  return (
+    <div className={cx(attributesName, align)} >
+      <div className={cx(bodyClassName)}>
```
StephenEsser: If `result-view-body` is not used anywhere the isPadded block above can be removed. This line can be updated to something along the lines of:
```
<div className={cx({ 'is-padded': isPadded })}>
```
This will add the is-padded class if isPadded is true, otherwise it will ignore it.

Here is some documentation on classNames: https://github.com/JedWatson/classnames#usage

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  let bodyClassName = 'result-view-body';
+
+  if (isPadded) {
+    bodyClassName = 'result-view-body--padded';
+  }
+
+  return (
+    <div className={cx(attributesName, align)} >
```
StephenEsser: Each of our components supports a concept of spreading customProps, this allows users to pass in any extra props to spread onto the outer tag of the component. Here is an example: https://github.com/cerner/terra-core/blob/master/packages/terra-card/src/Card.jsx

Any additional props are spread into a variable called customProps here: https://github.com/cerner/terra-core/blob/master/packages/terra-card/src/Card.jsx#L19

Merge any user provided classes into our own classes: https://github.com/cerner/terra-core/blob/master/packages/terra-card/src/Card.jsx#L23

Spread all additional custom props onto the outer most tag: https://github.com/cerner/terra-core/blob/master/packages/terra-card/src/Card.jsx#L26

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,164 @@
...
+const ResultView = (props) => {
+  const { results, timeDisplay, hasDocument, isModified, hasComment, alignment, isTruncated, isPadded } = props;
+  const attributesName = 'result-view';
+  let align = 'alignment-left';
+
+  if (alignment === 'right') {
+    align = 'alignment-right';
+  } else if (alignment === 'center') {
+    align = 'alignment-center';
+  }
```
StephenEsser:
```
  const attributesName = 'result-view';
  let align = 'alignment-left';

  if (alignment === 'right') {
    align = 'alignment-right';
  } else if (alignment === 'center') {
    align = 'alignment-center';
  }
```

This could be replaced with: 
```
const resultClassNames = cx('result-view', `alignment-${alignment}`);
```

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,164 @@
...
+          { resultsDiv(results, isTruncated) }
+          {isModified &&
+            <IconModified />
+          }
+          {hasComment &&
+            <IconComment />
+          }
+        </div>
+        {timeDisplay &&
+          timeDisplayDiv(timeDisplay)
```
StephenEsser: The timeDisplayDiv function is small enough it could probably be removed and just inline the html tags here. This makes it easy to read the markup.

```
{timeDisplay && <div className={cx('result-view-result-age')}>{timeDisplay}</div>}
```

packages/terra-clinical-result-view/src/ResultView.jsx
rakshith91: No, it's not used anywhere. I'll change it to null

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,164 @@
...
+
+    let unitDiv = null;
+    if (resultValue.unit && resultValue.unit.length > 0) {
+      unitDiv = <small className={cx('result-view-result-unit')} >{resultValue.unit}</small>;
+    }
+
+    const resultValueDisplay = resultValue.value || '--';
+    const resultValueTextClassName = isTruncated ? 'result-view-truncated' : 'result-view-result-value-text';
+
+    const resultValueDiv = (<span className={cx(resultValueTextClassName)} style={{ color: resultValue.color }} >
```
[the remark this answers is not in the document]
rakshith91: Changed all of them.

packages/terra-clinical-result-view/src/ResultView.jsx
```diff
@@ -0,0 +1,149 @@
...
+    isPadded,
+    ...customProps } = props;
+
+  const resultClassNames = cx('result-view', `alignment-${alignment}`, customProps.className);
+
+  return (
+    <div {...customProps} className={cx(resultClassNames)}>
+      <div className={cx({ 'result-view-body--padded': isPadded })}>
+        <div className={cx('result-view-result-block')}>
+          {hasDocument &&
```
yuderekyu: Nit: for readability, we can make these conditionals a single line
```jsx
{hasDocument && <IconDocuments />}
{isModified && <IconModified />}
...
```

dkasper-was-taken:
Truncation does not appear to work on firefox.
screen shot 2018-04-17 at 1 23 35 pm

rakshith91:
> Truncation does not appear to work on firefox.

fixed.

packages/terra-clinical-result-view/src/IconCritical.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
emilyrohrbough: Should these be in terra-icon?

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -26,7 +26,7 @@
   }
 
   .result-view-body--padded {
-    padding: var(--result-view-body-padding, 0.3125rem);
+    padding: var(--result-view-body-padding, 0.3125em);
```
emilyrohrbough: How come we switched to `em`?

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -35,25 +35,81 @@
...
   .result-view-indicator {
-    padding-right: var(--result-view-indicator-padding-right, 0.1875rem);
+    padding-right: var(--result-view-indicator-padding-right, 0.1875em);
+
+    &.critical {
+      color: var(--result-view-text-color, #c00);
+    }
+
+    &.low {
+      color: var(--result-view-low-color, #0053e6);
```
emilyrohrbough: These should probably be more specific to what they are themeing as well as name-spaced to the terra component to avoid potential name collisions : `--terra-clinical-result-view-normalcy-indicator-low`

emilyrohrbough: ^ It actually looks like this themable naming is reused in different places with fallback set a few times..

packages/terra-clinical-result-view/src/IconCritical.jsx
rakshith91: terra-icon has these icons with pre-filled colors. These ones are to make icons theme-able.

packages/terra-clinical-result-view/src/ResultView.scss
rakshith91: This component can be consumed in different places with different font sizes. ems would be a better choice like @neilpfeiffer suggested.

emilyrohrbough:
@rakshith91 What was the reason for the Do Not Merge again?

rakshith91:
> @rakshith91 What was the reason for the Do Not Merge again?

@neilpfeiffer is still working on the UX

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -0,0 +1,125 @@
+@import './mixins';
+
+:local {
+  .result-view {
+    color: var(--result-view-background-color, #595f62);
```
hboege: Could the variable please be renamed?  Color is for text in a control, not the background.

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  .result-view-result-unit {
+    color: var(--result-view-result-unit-text-color, #6f7477);
+    padding-left: var(--result-view-result-unit-padding-left, 0.1875em);
+    padding-right: var(--result-view-result-unit-padding-right, 0.1875em);
+
+    &.critical {
+      color: var(--result-view-result-unit-text-critical, #c00);
```
hboege: should the word 'color' be anywhere in the variable name?

packages/terra-clinical-result-view/src/ResultView.scss
```diff
@@ -0,0 +1,125 @@
...
+:local {
+  .result-view {
+    color: var(--result-view-background-color, #595f62);
+    font-size: var(--result-view-font-size, 1.25em);
+    line-height: var(--result-view-line-height, 1.25);
+    white-space: nowrap;
+    width: 100%;
+
+    svg {
+      fill: currentColor;
```
hboege: How is currentColor being set?   Can it be changed by a theme?

mjhenkes:
@ryanthemanuel @rakshith91  What is the status on feature? Should we close this PR?

ryanthemanuel:
From @neilpfeiffer "Sounds good. I think I can pick up the Clinical Result again starting Monday."  So hopefully this will regain traction again on 10/1

mjhenkes:
Closing due to inactivity.

PR #336, opened 2018-05-17
Consume terra-dev-site 1.0

mjhenkes:
Delete site, action header, error view and no data view

### Summary
Move examples.
Delete site, action header, error view and no data view

@cerner/terra


postcss.config.js
```diff
@@ -1,3 +1,3 @@
-const config = require('./packages/terra-clinical-site/postcss.config');
+/* eslint-disable import/no-extraneous-dependencies */
```
emilyrohrbough: The default config handle this, so this can be removed

mjhenkes: the whole file?

emilyrohrbough: Yes the postcss configuration is defined in terra-dev-site's webpack config

emilyrohrbough:
We can remove the i18n section on the repo readme at the time since there no internationalized components in terra-clincial

mjhenkes:
@emilyrohrbough onset picker is internationalized, but that section on the read me is dumb and gets outdated very easily. I'll remove it.

emilyrohrbough:
Ah, forgot about that one.

PR #351, opened 2018-06-12
Label Value View Theming

### LINKED ISSUE #344: LabelValueView needs theme variables
# Issue Description
In the current LabelValueView implementation, we cannot change the colors for a theme.  We need this for the lowlight theme, as just using default colors makes the text unreadable.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [X] Other

## Expected Behavior
Have variables for any colors hardcoded in the current scss for LabelValueView.

## Current Behavior
Colors cannot be changed per theme.

mayurikini:
### Summary
This change will allow the label text color to be themeable for LabelValueView component.

Resolves #344


packages/terra-clinical-label-value-view/src/LabelValueView.scss
```diff
@@ -7,7 +7,7 @@
   }
 
   .label {
-    color: #75767d;
+    color: var(--terra-label-value-label-color, #75767d);
```
dkasper-was-taken: Update the variable to --terra-label-value-view-label-color;
@neilpfeiffer Can you verify?

packages/terra-clinical-label-value-view/src/LabelValueView.scss
```diff
@@ -7,7 +7,7 @@
...
     font-size: 0.8571428571rem;
```
yuderekyu: Should we open up font-size, line-height, and margin-bottom to be themeable too?

mayurikini: Scope of this PR is limited to label text color.

yuderekyu: There's potential for other consumers to want to customize these values. Since this is the first run through to create themeable variables on clinical, it'd be helpful to discuss which values actually need to be themeable. Tagging @neilpfeiffer for input. If we do, we can log another issue so it's not a blocker.

PR #354, opened 2018-06-18
Update base dependencies

### LINKED ISSUE #347: Only define base as peer dependency
# Issue Description
Terra components have been declaring terra-base as a dependency and a peer dependency. We should move this to be a peer dependency to ensure only one version is installed and referenced.

See discussion here: https://github.com/cerner/terra-core/pull/1583/files

Sync https://github.com/cerner/terra-core/issues/1587

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

emilyrohrbough:
### Summary
Closes #347


packages/terra-clinical-application/package.json
```diff
@@ -33,7 +33,6 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "prop-types": "^15.5.8",
-    "terra-base": "^3.0.0",
```
StephenEsser: Do we want to bump the peerDependency to the latest terra-base version? Should lerna be doing that automatically?

emilyrohrbough: It appears lerna hasn't been bumping peer dependencies 😞

StephenEsser: I wonder if it will only bump the peer dependency if they are in the same monorepo

bjankord: I believe lerna does not handle bumping peerDeps. We have to do that manually. I think leaving this at `^3.0.0` is fine or setting it to `^3.12.0` is also fine. Both pull in the latest minor version of v3, its just a diff in what the lower bound version be. As long as the v3 minor and patch releases are following semver, using `^3.0.0` for the peerDep shouldn't be an issue.

PR #357, opened 2018-06-19
Fix DetailView title contents from overflowing and not breaking

### LINKED ISSUE #355: Title element in Detail View not breaking mid-word when inside smaller container
# Issue Description
The title element in the clinical detail view does not break its text mid-word.  It will only break the text to wrap when there is a space or '-'.  This causes the detail view to have a larger width than its container, ie when inside a slide panel.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
We would expect the text to break in the middle of the word.  

## Current Behavior
For a really long string given to the title prop, The string is not broken when the string is all one word and is wider than the width of it's container.

screen shot 2018-06-18 at 3 07 07 pm
vs
screen shot 2018-06-18 at 3 07 24 pm

tomleewu:
### Summary
Fixes issue where long string (without spaces) would overflow its container in `DetailView`.

Closes #355.

@cerner/terra


yuderekyu:
Nice - want to also add a continuous long text wdio test case to validate this?

yuderekyu:
A second thought, is this also a problem with the other sections? If so, we'll can add the `word-wrap` property to those to fix all overflow cases.

packages/terra-clinical-detail-view/src/terra-dev-site/test/clinical-detail-view/DetailView/DetailViewOverflowText.test.jsx
```diff
@@ -8,9 +8,9 @@ const accessoryElement = (<p> A sample accessory container </p>);
 
 const DetailViewOverflowText = () => (
   <DetailView
-    title="OverflowTextOverflowTextOverflowTextOverflowTextOverflowTextOverflowTextOverflowTextOverflowTextOverflowText"
-    secondaryTitles={['secondaryTitle1', 'secondaryTitle2']}
-    subtitles={['Subtitle 1', 'Subtitle 2']}
+    title={'OverflowText'.repeat(10)}
```
emilyrohrbough: Looking at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat it seems this maybe not be compatible on IE? Do this render as expected in these browsers?

tomleewu: They appear to be working

### IE10
screen shot 2018-06-19 at 4 35 04 pm

### IE11
screen shot 2018-06-19 at 4 33 38 pm

PR #358, opened 2018-06-19
Use Terra's Config Modules

emilyrohrbough:
### Summary
- Use `browerslist-config-terra`, `eslint-config-terra`, and `stylelint-config-terra` modules.
- Fix lint errors.


package.json
```diff
@@ -21,6 +21,22 @@
   "bugs": {
     "url": "https://github.com/cerner/terra-clinical/issues"
   },
+  "browserslist": [
```
emilyrohrbough: Unfortunately we cannot use 
```
"browserslist": [
   "extends browserslist-config-terra",
],
```
here because when we minimize our css during `npm run compile:prod`, the `css-loader`'s `cssnano` v3.10.0 / its dependencies pulls in `browserlist` v1.7.7. This version of browserslist is before the extends capability was introduced so webpack compile fails to build in prod. 

The quickest solution is to copy - and - paste our specified browsers. We must specify the browsers we use because `estlint-config-terra` provides a compatibility that uses this browserslist configuration and if we do not provide a config, it uses browserlist's default configuration of `> 0.5%, last 2 versions, Firefox ESR, not dead` !== our targeted browsers. 

The good news is the cssnano maintainer is in the process up updating their packages, but this will still require stable release + css-loader + deps to be updated. Here is the dependency chain found in the package-lock:
```
terra-toolkit                        3.9.0
css-loader                          0.28.14
cssnano                              3.10.0
post-css-merge-rules       2.1.2
caniuse-api                         1.6.6
browserslist                        1.7.7
```

Issue will be logged to update these once css-loader pulls the latest release of cssnano. @mjhenkes and I talked offline and it appears other css-minimization modules are also leveraging cssnano, so hard-coding this config is the 'easiest' win.

emilyrohrbough:
Blocked by #359

PR #368, opened 2018-07-05
Added color variables to Header

### LINKED ISSUE #366: Header needs theme variables for color attributes
# Feature Request
In the current Header implementation, we cannot change the colors for a theme. We need this for the lowlight theme, as just using default colors makes the text unreadable.

## Description
Add variables to theme header's and subheader's background and any other colors that might exist.

cwalten:
### Summary
Color variables have been added to the Header component. Resolves #366 


packages/terra-clinical-header/src/Header.scss
```diff
@@ -1,7 +1,7 @@
 :local {
   .flex-header {
-    background-color: #f4f4f4;
-    border-bottom-color: #dedfe0;
+    background-color: var(--terra-header-background-color, #f4f4f4);
```
emilyrohrbough: Can we updates these to be name-spaced to terra-clinical?

cwalten: Based on other scss files in this repo, I omitted clinical from the name. I have no problem adding the name spacing here, just wanted to make sure we are aware we are breaking that pattern.

emilyrohrbough: oh no... IMO  that pattern that should have been followed for the clinical components.
@ryanthemanuel @mjhenkes thoughts?

ryanthemanuel: Yes, let's go ahead make this one terra-clinical.  The other ones probably should be as well, but that would be non-passive at this point.

PR #369, opened 2018-07-05
Remove CSS 'initial' value and fix invalid value

Blackop778:
### Summary
`initial` values are not supported by Internet Explorer so it was changed to it's equivalent. `0` is not a valid value for `border-bottom-style` so it was changed to `none`.

Sync with cerner/terra-core#1667

@cerner/terra


packages/terra-clinical-header/src/Header.scss
```diff
@@ -64,8 +64,8 @@
 
   .flex-subheader {
     background-color: #c8cacb;
-    border-bottom-color: initial;
-    border-bottom-style: 0;
+    border-bottom-color: currentColor;
```
yuderekyu: Curious, if `border-bottom-color` is removed, do we still have the same inherited behavior on the border?

Blackop778: The border color does get set to the currentColor if border-bottom-color is removed.

Issue #378, opened 2018-07-24
Item Display Accessibility Issues


Blackop778:
# Bug Report

## Description
The `terra-clinical-item-display` ItemComment text color does not have enough contrast with the `terra-list` selected background color, the `terra-table` selected background color, or the `terra-table` striped background color.

## Steps to Reproduce
1. [Visit the selectable test page](https://engineering.cerner.com/terra-clinical/#/tests/terra-clinical-item-collection/clinical-item-collection/item-collection-selectable)
2. Test color contrast using Axe in table mode (larger viewport)
3. Test color contrast using Axe in list mode (smaller viewport)
4. [Visit the table styles test page](https://engineering.cerner.com/terra-clinical/#/tests/terra-clinical-item-collection/clinical-item-collection/item-collection-table-styles)
5. Test color contrast using Axe in table mode (larger viewport)

### Selectable
<img width="133" alt="screen shot 2018-07-24 at 2 43 56 pm" src="https://user-images.githubusercontent.com/11130785/43162282-26acdb1e-8f50-11e8-90a3-f45bad7f62f1.png">

```
expected no accessibility violations but got:
	{
  "description": "Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds",
  "help": "Elements must have sufficient color contrast",
  "helpUrl": "https://dequeuniversity.com/rules/axe/3.0/color-contrast?application=axeAPI",
  "id": "color-contrast",
  "impact": "serious",
  "nodes": [
    {
      "all": [],
      "any": [
        {
          "data": {
            "bgColor": "#e2f2fc",
            "contrastRatio": 3.94,
            "expectedContrastRatio": "4.5:1",
            "fgColor": "#75767d",
            "fontSize": "10.5pt",
            "fontWeight": "normal"
          },
          "id": "color-contrast",
          "impact": "serious",
          "message": "Element has insufficient color contrast of 3.94 (foreground color: #75767d, background color: #e2f2fc, font size: 10.5pt, font weight: normal). Expected contrast ratio of 4.5:1",
          "relatedNodes": [
            {
              "html": "<li tabindex=\"0\" id=\"item3\" class=\"List-module__list-item___1QAkf List-module__selected___rDzz4 List-module__is-selectable___38JEl\">",
              "target": [
                "#item3"
              ]
            }
          ]
        }
      ],
[...]

PR #379, opened 2018-07-24
[item-collection] Fix list role being missing

### LINKED ISSUE #364: role="listbox" attribute missing from <ul> within terra-item-collection component
# Bug Report

## Description
[Originally logged in terra-core](https://github.com/cerner/terra-core/issues/1656) by @juliacalandro 

When we upgraded to version 2.7.0 of terra-list, we received the following accessibility violation from our wdio tests:

_"description": "Ensures elements with an ARIA role that require parent roles are contained by them",
  "help": "Certain ARIA roles must be contained by particular parents",
  "helpUrl": "https://dequeuniversity.com/rules/axe/2.6/aria-required-parent?application=axeAPI",
  "id": "aria-required-parent",
..................
"nodes": [
    {
      "all": [],
      "any": [
        {
          "data": [
            "listbox"
          ],
          "id": "aria-required-parent",
          "impact": "critical",
          "message": "Required ARIA parent role not present: listbox",
          "relatedNodes": []
        }
      ],
      "failureSummary": "Fix any of the following:\n  Required ARIA parent role not present: listbox"_

After inspecting the terra-list code, we noticed that the `role="option"` attribute was getting set on `<li>` within the ListItem component, but a role attribute was not getting set on `<ul>` within the List component. I looked into this issue a bit further, and found [this webpage](http://w3c.github.io/aria-practices/examples/listbox/listbox-scrollable.html) about aria/accessibility which states that `role="listbox"` needs to be an attribute on `<ul>` which is the parent of `<li>`.

## Steps to Reproduce
Install the aXe extension for Chrome.
Run a project using terra-list so that it can be opened and inspected in the browser.
Inspect the page and navigate to where the List component is being used.
Verify that `<ul>` does not have a role set but its child `<li>`s do.
Navigate to the aXe tab and run an inspection on the page.
Verify that an accessibility error is thrown with a description matching the error description above.
Add `role="listbox"` as an attribute to each offending `<ul>` component and run the aXe inspection again.
Verify that the accessibility issue doesn't exist anymore.

## Additional Context / Screenshots
screen shot 2018-06-27 at 11 22 05 am

## Expected Behavior
I expect the accessibility issue mentioned above to no longer be thrown by the terra-list component.

## Possible Solution
Add missing `role="listbox"` attribute to `<ul>` within terra-list component.

## Environment
terra-list 2.6.0
Google Chrome Version 67.0.3396.99 (Official Build) (64-bit)

Blackop778:
### Summary
Item collection lists will have a role tag applied if any of it's children have a chevron (which is the ListItem logic for adding the option role that was causing the issues).

Also adds accessibility testing for the item collection. The found accessibility issue is listed at #378.

### Additional Details
Fixes #364

@cerner/terra


packages/terra-clinical-item-collection/src/_ListView.jsx
```diff
@@ -39,10 +39,13 @@ function createListItems(children, onSelect, requiredElements) {
 const ListView = ({
   children, onSelect, isListDivided, requiredElements,
 }) => {
+  const hasAChevron = React.Children.toArray(children).some(child => child.props.showListItemChevron);
+
   const listItems = createListItems(children, onSelect, requiredElements);
+  const roleSpread = hasAChevron ? { role: 'listbox' } : {};
```
emilyrohrbough: Why are we only applying this role when `hasAChevron` is true?  Should we not apply this at all times?

Blackop778: If the list has `role="listbox"` but doesn't have a child with `role="option"` then that is considered an accessibility violation. `hasAChevron` will be true only if at least one child of the list has `role="option"`.

emilyrohrbough: I'm not super familiar with list accessibility rules, so should we be adding `role="option"` to each list item regardless of the chevron?

Blackop778: According to the [specification for the option role](https://www.w3.org/TR/2010/WD-wai-aria-20100916/roles#option) it is for
> A selectable item in a select list.

so I suppose we should be checking for the `isSelectable` prop here and in `terra-list` instead.

packages/terra-clinical-item-collection/src/_ListView.jsx
```diff
@@ -39,7 +39,7 @@ function createListItems(children, onSelect, requiredElements) {
 const ListView = ({
   children, onSelect, isListDivided, requiredElements,
 }) => {
-  const hasAChevron = React.Children.toArray(children).some(child => child.props.showListItemChevron);
+  const hasAChevron = React.Children.toArray(children).some(child => child.props.isSelectable);
```
emilyrohrbough: Can we update this const to reflect we are checking for isSelectable to toggle the role assigned below?

packages/terra-clinical-item-collection/src/_ListView.jsx
```diff
@@ -39,10 +39,13 @@ function createListItems(children, onSelect, requiredElements) {
...
   children, onSelect, isListDivided, requiredElements,
 }) => {
+  const hasSelectableChild = React.Children.toArray(children).some(child => child.props.isSelectable);
+
   const listItems = createListItems(children, onSelect, requiredElements);
+  const roleSpread = hasSelectableChild ? { role: 'listbox' } : {};
 
   return (
-    <List data-terra-clinical-item-collection-list-view isDivided={isListDivided} >
+    <List data-terra-clinical-item-collection-list-view isDivided={isListDivided} {...roleSpread}>
```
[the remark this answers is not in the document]
Blackop778: Jest only snapshots the React elements and their props that would be rendered so it wouldn't have noticed the attributes on the HTML element changing.

PR #380, opened 2018-07-25
Bump responsive element version

bjankord:
### Summary
Missed bumping this to new major version in last update. This corrects the version so using the latest release of terra-ui components will all use terra-responsive-element v3.

### DIFF
--- a/packages/terra-clinical-item-collection/package.json
+++ b/packages/terra-clinical-item-collection/package.json
@@ -32,7 +32,7 @@
     "terra-doc-template": "^1.2.0",
     "terra-icon": "^2.9.0",
     "terra-list": "^2.0.1",
-    "terra-responsive-element": "^2.0.1",
+    "terra-responsive-element": "^3.0.0",
     "terra-table": "^2.0.1"
   },
   "scripts": {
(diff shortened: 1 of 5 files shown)


packages/terra-clinical-item-collection/package.json
```diff
@@ -32,7 +32,7 @@
     "terra-doc-template": "^1.2.0",
     "terra-icon": "^2.9.0",
     "terra-list": "^2.0.1",
-    "terra-responsive-element": "^2.0.1",
+    "terra-responsive-element": "^3.0.0",
```
StephenEsser: I think this change might be none passive. We were previously spreading custom props onto the responsive element which is not part of the API in version 3.0. I think we'll need to wrap the responsive element in a div to stay passive:

https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-item-collection/src/ItemCollection.jsx#L118


Something along the lines of:
```
<div {...customProps}>
    <ResponsiveElement
      responsiveTo="parent"
      defaultElement={listDisplay}
      {...breakpointDisplay}
    />
</div>
```

bjankord: Adding the wrapping div and a snapshot test case to capture that custom props are passed through correctly. https://github.com/cerner/terra-clinical/pull/380/commits/a7df384d65d18b52717c0b4e8d6786229bb8877c

PR #408, opened 2018-10-22
DataGrid Improvements

tbiethman:
### Summary
A couple things changed here:

* When the 'fill' prop is not set, the horizontal scroll synchronization and custom horizontal scrollbar are unnecessary. I've updated the markup in that case to remove those elements and make things more efficient.
* The row striping was using an nth-child selector, which could be influenced by the presence of a section header. I've made the row striping API explicit and implemented it on a per-section basis, so rows should be more visually consistent.
* A few event handlers were being set when they didn't need to be. I added checks around them to make sure only only the necessary handlers are being executed.
* In some circumstances, the header cells were rendering with a 2px offset (due to the presence of the bottom border on their container). I moved the border styles around to ensure proper alignment.

Not much has changed functionally, but the wdio screenshots got hit pretty hard due to the row striping and header cell offset changes.


packages/terra-clinical-data-grid/src/subcomponents/Row.module.scss
```diff
@@ -1,11 +1,10 @@
...
   .row {
     background-color: var(--terra-clinical-data-grid-row-background-color, #fff);
-    display: flex;
     min-width: 100%;
   }
 
-  .row:nth-child(odd) {
+  .row.striped {
     background-color: var(--terra-clinical-data-grid-row-striped-background-color, #f0f0f0);
   }
```
benbcai: The variable `--terra-clinical-data-grid-row-hover-background-color` is currently set to have two different default values:

>   .row.hover {
    background-color: var(--terra-clinical-data-grid-row-hover-background-color, #e2f2fc);
  }

> .row.selected.hover {
    background-color: var(--terra-clinical-data-grid-row-hover-background-color, #b3ddf7);
  }

Perhaps, rename the latter to `--terra-clinical-data-grid-row-selected-hover-background-color` ?

tbiethman: +2 nice find, I will update that.

packages/terra-clinical-data-grid/src/DataGrid.module.scss
```diff
@@ -10,42 +10,48 @@
```
benbcai: The variable `terra-clinical-data-grid-pinned-column-box-shadow` is used in two different places with different default styles (one has an inset).

tbiethman: Double checked with Neil, the separate values are intended. So I renamed the variable for the scrollbar to `--terra-clinical-data-grid-scrollbar-pinned-column-box-shadow`

PaulSolDev:
Functional verification complete.  Data Grid appears to be working as designed with the new enhancements.  Significant performance improvement noticed for vertical scroll behavior on IE compared with other implementations.  Horizontal scroll remains delayed.  Hover behavior for the cells is slightly delayed as well (IE only). No new issues identified.

PR #417, opened 2018-11-01
Generate TAR archive files for individual packages

### LINKED ISSUE #416: Generate TAR archive files for individual packages
# Feature Request
Generate tar archive files for individual packages. 

This will allow theme repositories to reference packages directly from the mono repo and mitigate the performance hit taken by installing the entire monorepo.

## Description

Introduce an additional step to the compile script to generate a new build artifact for each package. These packages will need to be checked in, but will not be released with the package. This can be enforced by upgrading the danger script to detect changes to source files include a change to the tar archive.

StephenEsser:
### Summary
- Modified each package's compile script to generate a new tar archive
- Modified the danger script to require any change to a source file to be accompanied by a new tar file
- Added the new tar archive to the npm ignore 

### Additional Details
This will allow theme repositories to reference packages directly from the mono repo and mitigate the performance hit taken by installing the entire monorepo.

Resolves #416 

@cerner/terra


packages/terra-clinical-application/package.json
```diff
@@ -36,7 +36,7 @@
     "terra-doc-template": "^1.2.0"
   },
   "scripts": {
-    "compile": "npm run compile:clean && npm run compile:build && rm terra-clinical-application.tgz && npm pack --ignore-scripts && mv *.tgz terra-clinical-application.tgz",
+    "compile": "npm run compile:clean && npm run compile:build && tt-pack",
```
emilyrohrbough: Just a side node, I am a little worried about naming this script `tt-pack` because we have been telling people to add a `pack` script to projects in referencing to webpack packing assets.

StephenEsser: I'm writing a note about it on the terra-toolkit PR, we can move discussions over on that when it is opened. 

The challenge is this script is essentially running `npm pack`. https://docs.npmjs.com/cli/pack

emilyrohrbough: Yeah... I want our terminology to match npm but we have also made it weird with out script naming preference for webpack. 🤷‍♀️ Just wanted to mention something, mostly so I wouldn't' forget, but lets move discussion there.

Issue #396, opened 2018-09-19
Onset Picker crashing


PaulSolDev:
# Bug Report

## Description

Date field in Onset Picker is causing a crash when the value is cleared.

## Steps to Reproduce

1. Select 'Date' option in the onset picker
2. Clear the date field 
3. Browser page appears to crash/ go blank

## Additional Context / Screenshots

![image](https://user-images.githubusercontent.com/29487301/45773977-692fb400-bc11-11e8-8b38-e5cf47c3776c.png)

## Expected Behavior

Expect date value to clear and for the empty field to support entry of a new date value.

## Environment

* Component Name and Version: Onset Picker
* Browser Name and Version: All browsers
* Node/npm Version: [e.g. Node 8/npm 5]
* Operating System and version (desktop or mobile): Tested on Desktop

ryanthemanuel:
@doogthemushroom is going to pick this up.

emilyrohrbough:
Resolved in https://github.com/cerner/terra-clinical/pull/427

Issue #256, opened 2018-01-25
Uplift Onset Picker to utilize new Form Selects


doogthemushroom:
# Issue Description
Would like to see the current Onset Picker uplifted to use new Form Selects utilizing OCS. Would also like to see an option to change the default value from current date to unselected as well.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
I expect my form to look slick with OCS and this OnsetPicker is a real drag, man! Expect to use Form Select from terra-form-select.

## Current Behavior
Uses Select from terra-form.

ryanthemanuel:
@doogthemushroom is going to pick this up.

ryanthemanuel:
@doogthemushroom do we have an eta on when this will get done?  I think on 11/29, you said it was going to be out for a PR soon.

ryanthemanuel:
Spoke with @doogthemushroom offline and he said he should have the PR out by EOD 12/3.

doogthemushroom:
I think this can be closed as a result of #427

PR #427, opened 2018-12-04
Uplift Onset Picker form inputs/selects, fix crash (#256, #396)

doogthemushroom:
### Summary
These changes aim to update the Onset Picker to use the new Terra Form components (#256) and address the crash found in #396. The name prop values have also been condensed into a single prop that is used as a base value to create unique ids and names for child components. These changes will necessitate a major version bump.

### Additional Details
The new form components support blanks, and resetting the date was the requested behavior in #396, so a number of changes were done to handle undefined/blank values. ~~New translations were added but are currently all English. I will hopefully have correct translations by the end of the week.~~
Translations are now up to date, but lacking arabic which cannot be added at this time. The readme has been updated to reflect this.

### Before merge
Remove the tar file!


packages/terra-clinical-onset-picker/src/OnsetPicker.scss
```diff
@@ -12,4 +12,26 @@
     margin-bottom: 0.357em;
```
ryanthemanuel: Let's make sure and create themeable variables for these values.  Take a look at other components for what/how we typically theme.

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
```diff
@@ -12,4 +12,26 @@
...
+
+  .month {
+    min-width: 150px;
+  }
+
+  .age-unit {
+    min-width: 100px;
+  }
+
+  @media screen and (min-width: 544px) {
```
tbiethman: The terra-breakpoints package has media queries that you can import: https://github.com/cerner/terra-core/blob/master/packages/terra-breakpoints/src/_media-queries.scss#L1

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -38,17 +38,17 @@ const propTypes = {
...
   /**
    * A callback function to execute when a granularity is selected.
    * The first parameter is the changed granularity value.
    */
   granularitySelectOnChange: PropTypes.func,
 
+  /**
+   * The id of the onset picker. Used as the base for other required id/name in sub-components.
+   */
+  id: PropTypes.string.isRequired,
```
tbiethman: Typically the addition of a new required prop would be considered non-passive. Do we want to bump the version, or is the consumption base so low that we're comfortable not doing so?

mjhenkes: If a new required prop is added this will have to be major versioned.

ryanthemanuel:
How easy would it be to separate out the crash fix from the rest of this work?  I'm hesitant to tie them together.

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
doogthemushroom: Do you want just these margins to be made into variables? Or should I also include the min-width values?

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
doogthemushroom: Ah yes the plan was to go to the next major version, I'll update that

doogthemushroom:
@ryanthemanuel It should be fairly simple to separate the crash fix. Those changes were fairly minor.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -38,17 +38,17 @@ const propTypes = {
     GranularityOptions.DATE,
```
yuderekyu: If we are doing a major version bump, I vote to change the exported enums to strings. 
### Before
``` jsx
<OnsetPicker granularity={OnsetPicker.Opts.Granularities.AGE} precisionSet={OnsetPicker.Opts.Precisions.ONAT} />
```
### After
``` jsx
<OnsetPicker granularity="age" precisionSet="on/at" />
```
We had discussion about this here: https://github.com/cerner/terra-core/issues/1044

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
```diff
@@ -12,4 +12,26 @@
...
+  .precision {
+    min-width: 150px;
+  }
+
+  .granularity {
+    min-width: 100px;
+  }
+
+  .month {
+    min-width: 150px;
```
benbcai: When the granularity is Age and you switch between selecting Weeks and Months as the age unit, the width of the age value input changes.

doogthemushroom: Yeah I just noticed this, I'll be adding a fixed width to the age input

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -281,59 +269,89 @@ class OnsetPicker extends React.Component {
...
         />
       );
 
       ageUnitSelect = (
         <SelectField
-          className={cx('field-inline')}
-          data-terra-clinical-onset-picker="age_unit"
-          options={OnsetUtils.allowedAgeUnits(this.props.birthdate, intl)}
-          defaultValue={this.state.ageUnit.toString()}
+          className={cx('field-inline', 'age-unit')}
```
yuderekyu: Can we create reusable logic here? `ageUnitSelect`, `monthSelect`, and `yearSelect` are almost one-one. 

We could create reusable function to render a `<SelectField />`. This function would accept prop values `defaultValue`, `value`, `onChange` and `selectId`. Lastly, a parameter `selectType` to represent "age", "month", or "year". Check this param to set classNames, translations, and the`<SelectField.Option />` children components. 

Thoughts?

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
doogthemushroom: It seems the sass loader is having issues finding that file to import after I include the package. Is there a change in the webpack configuration that needs to happen first for that sass import to work properly?

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
doogthemushroom: Updated to 3.0.0

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
doogthemushroom: Made the margins and widths themeable variables. I also converted the widths to `em`'s

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
```diff
@@ -9,7 +9,33 @@
   }
 
   .field-inline {
-    margin-bottom: 0.357em;
-    margin-right: 0.357em;
+    margin-bottom: var(--terra-onset-picker-field-margin-bottom, 0.357em);
```
yuderekyu: let's add the `clinical` namespace to the themeable variables.
```suggestion
    margin-bottom: var(--terra-clinical-onset-picker-field-margin-bottom, 0.357em);
```

emilyrohrbough:
@doogthemushroom We will need an Upgrade Guide since this is a MVB

packages/terra-clinical-onset-picker/tests/wdio/onset-spec.js
```diff
@@ -1,5 +1,6 @@
 const viewports = Terra.viewports('tiny', 'small', 'medium', 'large', 'huge', 'enormous');
-const rules = { label: { enabled: false } };
+// Color contrast rule disabled due to https://github.com/cerner/terra-core/issues/1917
```
emilyrohrbough: Can we log a terra-clinical issue to remove this disabled accessibility rule once `https://github.com/cerner/terra-core/issues/1917` is resolved? 

Also, do we still need to disable the label accessibility rule? I don't think we should with the new form dependencies

doogthemushroom: Oh do you think the form tests would be covering the accessibility testing required for this component? Yeah I think I could agree with that.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
doogthemushroom: So pretty much just remove the exported enums and update docs? Easy enough.

packages/terra-clinical-onset-picker/tests/wdio/onset-spec.js
doogthemushroom: Oh whoops misread your comment. Yeah I'll check enabling the label rule.

doogthemushroom: It seems that the InputField is failing the label accessibility check. I'm not sure why, there is an associated label getting generated.

emilyrohrbough: hmmm @JakeLaCombe do you know what we might be missing in implementing the form input? I checked the package and there are no accessibility issues

doogthemushroom:
I misspoke about the ease of separating the crash from this update. I'm going to leave them together. It comes down to the deprecated Select's not supported blank values. This results in an inconsistency between the onsetDate getting reset when clearing the date in the Date Picker, and then the incorrect displayed values in the Onset Picker Select's.

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
doogthemushroom: Added clinical namespace.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
doogthemushroom: The only same thing is the SelectField and the isLabelHidden prop. Everything else would need to be different (albeit slightly) and thus included in the switch statement for the different select.  Doesn't seem like it would save many LOC or be that much clearer.

doogthemushroom:
@emilyrohrbough Added an Upgrade guide to the docs folder.

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
bjankord: Try adding this to the top of the file `@import '~terra-breakpoints/lib/media-queries';`
If that doesn't work, let me know and we can debug.

doogthemushroom: Ah that import seemed to have worked... might want to update the docs for that project!

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
```diff
@@ -1,4 +1,16 @@
+@import '~terra-breakpoints/lib/media-queries';
```
yuderekyu: let's add modules before the file extension. `OnsetPicker.module.scss`.

more discussion about that here: https://github.com/cerner/terra-core/issues/1511

packages/terra-clinical-onset-picker/tests/wdio/onset-spec.js
```diff
@@ -1,5 +1,6 @@
 const viewports = Terra.viewports('tiny', 'small', 'medium', 'large', 'huge', 'enormous');
```
yuderekyu: Is there a reason for every single viewport? This component only responsively changes in tiny viewports, yet 288 tests run. To cover the same area, we could cut this to tiny, medium, enormous

packages/terra-clinical-onset-picker/src/OnsetPicker.scss
doogthemushroom: module extension added

packages/terra-clinical-onset-picker/tests/wdio/onset-spec.js
doogthemushroom: Changed viewports to tiny, medium, and enormous.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
tbiethman: You don't need to bump the version yourself, that'll happen during the actual release. You just need to add a `### Breaking Change` header to your changes in the CHANGELOG so that the person doing the release knows to bump the major version.

packages/terra-clinical-onset-picker/tests/wdio/onset-spec.js
doogthemushroom: Seems the hidden labels prop does not add aria labels needed to pass accessibility in the form field. I'll address this by adding the proper labels and log an enhancement to terra-core to address this.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
doogthemushroom: Ah alright, I've reverted it to 2.15.0 and included a breaking change header to the changelog.

doogthemushroom:
Translations have been updated.

packages/terra-clinical-onset-picker/README.md
```diff
@@ -5,6 +5,7 @@
 [![Build Status](https://badgen.net/travis/cerner/terra-clinical)](https://travis-ci.org/cerner/terra-clinical)
 
 The terra-clinical-onset-picker component provides users a way to enter or select an approximate date for onset scenarios.
+NOTE: This component does not support arabic translations.
```
emilyrohrbough: It might be good to call this out larger under an `### I18n Support` section maybe?

packages/terra-clinical-onset-picker/docs/UPGRADEGUIDE.md
```diff
@@ -0,0 +1,31 @@
+# Terra Clinical Onset Picker Upgrade Guide
+
+## Changes from 2.0 to 3.0
+
+### Breaking Changes
+
+* New required prop `id`
+* Removed exported enums `OnsetPicker.Opts.Granularities` and `OnsetPicker.Opts.Precisions`
+* Removed default prop values for `granularity` and `precision`
```
emilyrohrbough: What do people us instead of these prop options?

packages/terra-clinical-onset-picker/src/terra-dev-site/test/clinical-onset-picker/FiveYears.test.jsx
```diff
@@ -10,8 +10,8 @@ class FiveYears extends React.Component {
     moment.now = () => new Date(2016, 8, 20);
 
     this.state = {
-      precision: OnsetPicker.Opts.Precisions.BEFORE,
-      granularity: OnsetPicker.Opts.Granularities.YEAR,
+      precision: 'before',
+      granularity: 'year',
```
emilyrohrbough: I thought these were removed?

packages/terra-clinical-onset-picker/docs/UPGRADEGUIDE.md
doogthemushroom: The prop is still there, it's just the default value has been removed. I suppose the default is technically undefined now.

packages/terra-clinical-onset-picker/src/terra-dev-site/test/clinical-onset-picker/FiveYears.test.jsx
doogthemushroom: Again, it was just the default values that were removed. Can still set the props.

packages/terra-clinical-onset-picker/README.md
doogthemushroom: Updated readme's to call this out.

PR #436, opened 2018-12-20
Improve site accessibility

jmsv6d:
### Summary
Addresses site accessibility issues for https://github.com/cerner/terra-ui/issues/46

@cerner/terra

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,14 +16,19 @@
 
 - [Supported Browsers](https://github.com/cerner/terra-ui/blob/master/src/terra-dev-site/contributing/ComponentStandards.e.contributing.md#cross-browser-support)
 - [Packages](#packages)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -49,7 +51,9 @@
 ## Internationalization (I18n)
 Please review [Terra's Internationalization documentation](https://engineering.cerner.com/terra-ui/#/getting-started/terra-ui/internationalization) for more information. Included are directions on consumption and how internationalization is setup.
 
-## Contributing
+<h2 id="contributing">
```
emilyrohrbough: Do we need to do this for Status and Internationalization as well?

jmsv6d: We only need ids for the headers that we have anchors to. But we should probably have anchors to those as well, I'll update.

PR #448, opened 2019-01-04
Changed All Component Terra-Core Dependency Versions

amichaelparker:
### Summary
Changed the dependency versions of all terra-core components to match hoists to remove lerna warnings for those components.

Closes [#400](https://github.com/cerner/terra-framework/pull/400)


packages/terra-clinical-data-grid/package.json
```diff
@@ -37,13 +37,13 @@
     "prop-types": "^15.5.8",
     "react-draggable": "^3.0.5",
     "resize-observer-polyfill": "^1.4.1",
-    "terra-button": "^2.0.1",
```
amichaelparker: This leaves react-intl as the only lerna warning (2.7.0 vs 2.4.0).

emilyrohrbough: Lets go ahead and match those as well. It looks like onset-picker is using 2.4.0. We can bump it to 2.7.0 to match data-grid or fall back to 2.0.0

amichaelparker: It was set to 2.7.0, and falling back to 2.0.0 also gives a hoist warning. Setting it to 2.4.0 seems to work.

emilyrohrbough: We would need to bump onset picker to 2.7.0 to match the data-grid's specification. I"m good with either way; the latest will be installed regardless.

amichaelparker: I went ahead and set both onset picker and data grid to 2.7.0: [b34fb92](https://github.com/cerner/terra-clinical/pull/448/commits/b34fb9201be7da2bdc94c4c4733f804eb77a545d)

yuderekyu:
sorry, we stealthily did most of this in the theme PR 😅

amichaelparker:
It happens... Pretty regularly. :D It looks like data-grid and item-view are still having problems with the icon version. Setting it to ^2.0.0 doesn't resolve it anymore, but ^2.9.0 does, so now I'm testing that to check for issues.

packages/terra-clinical-data-grid/package.json
```diff
@@ -37,13 +37,13 @@
...
     "terra-clinical-item-view": "^2.16.0",
-    "terra-content-container": "^2.0.0",
-    "terra-doc-template": "^1.0.0",
-    "terra-icon": "^2.0.0",
-    "terra-mixins": "^1.0.0",
-    "terra-overlay": "^2.0.0"
+    "terra-content-container": "^2.0.1",
+    "terra-doc-template": "^2.0.0",
+    "terra-icon": "^2.6.0",
+    "terra-mixins": "^1.21.0",
```
emilyrohrbough: uh oh, looks like we lost some of the changes you originally made

amichaelparker: Should I change those back? The new changes from master don't throw the same hoist warnings.

emilyrohrbough: yeah, we should match all dependencies to a major version `^2.0.0` or `^3.0.0` etc

emilyrohrbough: it will make it easier to manage if/when we add new components to the repo

amichaelparker: Packages updated!

packages/terra-clinical-detail-view/package.json
```diff
@@ -27,11 +27,10 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
-    "prop-types": "^15.5.8",
```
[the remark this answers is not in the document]
amichaelparker: Yeah, accidentally hacked that off while fixing merge conflicts.😀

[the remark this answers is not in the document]
amichaelparker: I also forgot to bump one back down to the nearest major version, so I'm on a kind of roll this morning.

emilyrohrbough:
A few changes: (I think it got lost in a merge)
- update `terra-doc-template` dependency to be `^2.0.0` [here](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-application/package.json#L36)
- update `terra-clinical-item-display` to be `^3.0.0` [here](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-item-view/package.json#L32) and [here](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-label-value-view/package.json#L32)

PR #452, opened 2019-01-11
Enable a docker dev environment

mjhenkes:
### Summary
This is a test to pull in the proposed docker image for development. 

Also the docker compose file is an example of what we'd have to put in each repo.

@cerner/terra


docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+    build:
+      context: ./
+      args:
+        - npm_install=false
+    ports:
+      - 8080:8080
+    volumes:
+      - .:/opt/module
+  standalone-chrome:
+    image: 'selenium/standalone-chrome:3.11.0-californium'
```
emilyrohrbough: Where you still going to create a terra selenium-docker image to dry this up? 

If we do, would need to have an TZ options for downstream passivity in consuming projects

mjhenkes: As of now, no I'm not drying this up with a terra selenium docker image. I'm not convinced the added maintenance is worth drying up 3 lines of code.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # use terra-node as base image
-FROM cerner/terra-node
+FROM mjhenkes/terra-base:1
```
mjhenkes: This is temporary until terra-toolkit gets released.

emilyrohrbough:
Once this is merge, I will make changes to https://github.com/cerner/terra-clinical/pull/454 to remove `shx` and the current script setup since we will not need to support windows syntax and can rely on linux commands.

PR #454, opened 2019-01-16
Repo Dependency / Script Changes

emilyrohrbough:
### Summary

- removed clean modules script in scripts/clean-modules
- removed post install script in scripts/post-install
    - no longer need the repo to be setup as a package
- removed `postinstall:package` and `postinstall:project` scripts in package.json
    - updated `postinstall` to be the `postinstall:project` script
- removed `bootstrap` script. We always want to use hoist while bootstrapping 
- replaced `shelljs` and `rimraf` to use `rm rf`.
- updated package-level clean scripts to use `rm rf` and drop `rimraf`


emilyrohrbough:
@mjhenkes this has been updated with your latest docker changes

emilyrohrbough:
@yuderekyu  We don't need to worry about windows compatibility issues for development setup since Matt dockerized the project. That means we can drop the extra dependencies

Issue #457, opened 2019-01-30
Provide attribute on DataGrid section header


jeffmerten:
# Feature Request

## Description
Provide an attribute on the section header of data grid.

## Additional Context / Screenshots
Our current solution has a navigation workflow where the user taps on a cell that effectively represents that section of the DataGrid. We'd like to add an attribute to the section header so we can scroll to that section on component mount.

## @ Mentions
@tbiethman

tbiethman:
Fixed by #458

PR #458, opened 2019-02-04
Add section id attribute on DataGrid headers

jeffmerten:
### Summary
Adds an attribute on the section header of terra-clinical-data-grid that is unique to the section id.

### Additional Details
For one of our use cases, we'd like to scroll the header into view on component mount. This change lets us properly do this.

### Related issue
https://github.com/cerner/terra-clinical/issues/457

### DIFF
--- a/packages/terra-clinical-data-grid/src/subcomponents/SectionHeader.jsx
+++ b/packages/terra-clinical-data-grid/src/subcomponents/SectionHeader.jsx
@@ -113,7 +113,7 @@ class SectionHeader extends React.Component {
     return (
       <div
         className={cx('section-header')}
-        data-terra-clinical-data-grid-section-header
+        data-section-header
         data-terra-clinical-data-grid-section-header-id={sectionId}
       >
         <div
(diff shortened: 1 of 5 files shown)


tbiethman:
> We should name space this data attribute:
> `data-terra-clinical-data-grid-section-header`

I think this makes sense. I'm not really worried about collisions, but fully scoping it might make this attribute feel more like a first-class part of the DataGrid's consumer API.

emilyrohrbough:
Yeah, I think its important ensure date attributes are named spaced just in case the possibility for collision. Especially because these attributes are added for callbacks  or for tests and changing them would be non-passive. List (new) and table (new) and toggle-section-head are components that all implement a section header.

packages/terra-clinical-data-grid/src/subcomponents/SectionHeader.jsx
```diff
@@ -113,7 +113,8 @@ class SectionHeader extends React.Component {
     return (
       <div
         className={cx('section-header')}
-        data-section-header
```
tbiethman: I would leave this one the way it is in the interest of passivity. Might be a little awkward but we can revisit it in the future if we need to (or just remove it outright if it's not being used).

jeffmerten: I didn't see anyone using it which is why I changed it.

tbiethman: We're not using it internally, but a theoretical consumer might be using it for some other purpose, and changing it would be non-passive for them.

PR #460, opened 2019-02-07
Fix lint errors

bjankord:
### Summary
Fix up some lint errors introduced with new version of eslint-react-plugin.

### DIFF
--- a/packages/terra-clinical-data-grid/src/subcomponents/Cell.jsx
+++ b/packages/terra-clinical-data-grid/src/subcomponents/Cell.jsx
@@ -11,6 +11,10 @@ const cx = classNames.bind(styles);
 
 const propTypes = {
   /**
+   * Accessible text provided for the Cell.
+   */
+  ariaLabel: PropTypes.string.isRequired,
+  /**
    * String identifier of the section in which the Cell will be rendered.
    */
   sectionId: PropTypes.string.isRequired,
--- a/packages/terra-clinical-data-grid/src/subcomponents/Cell.jsx
+++ b/packages/terra-clinical-data-grid/src/subcomponents/Cell.jsx
@@ -13,7 +13,7 @@ const propTypes = {
   /**
    * Accessible text provided for the Cell.
    */
-  ariaLabel: PropTypes.string.isRequired,
+  ariaLabel: PropTypes.string,
   /**
    * String identifier of the section in which the Cell will be rendered.
    */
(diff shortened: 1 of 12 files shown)


packages/terra-clinical-data-grid/src/terra-dev-site/test/clinical-data-grid/PagedContentDataGrid.test.jsx
```diff
@@ -94,7 +94,7 @@ class PagedContentDataGrid extends React.Component {
           sections={this.state.sections}
           fill
           onRequestContent={this.state.sectionCount < 2 ? (() => {
-            this.setState(prevState => ({ sectionCount: prevState.sectionCount + 1, isLoading: false, sections: PagedContentDataGrid.buildSections(prevState.sectionCount + 1) }));
```
bjankord: `isLoading` was not used elsewhere.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -271,14 +271,11 @@ class OnsetPicker extends React.Component {
     const {
       birthdate,
       granularity,
-      granularitySelectName,
```
bjankord: These were not used.

packages/terra-clinical-data-grid/src/subcomponents/Cell.jsx
```diff
@@ -10,6 +10,10 @@ import styles from './Cell.module.scss';
 const cx = classNames.bind(styles);
 
 const propTypes = {
+  /**
+   * Accessible text provided for the Cell.
+   */
+  ariaLabel: PropTypes.string.isRequired,
```
StephenEsser: Is this always passed in?

bjankord: Not 100% sure, my assumption was that is was but maybe I should drop the `.isRequired` part.

tbiethman: It shouldn't be required, only the RowSelectionCell provides it because it's the only Cell without custom content. Any other cell would provide aria attributes through its custom content.

bjankord: Updated in this commit: https://github.com/cerner/terra-clinical/pull/460/commits/10a6eafbc38572be48d4071d98ee59642aa7c840

Issue #461, opened 2019-02-11
Onset Picker does not include legend with fieldset


bjankord:
# Bug Report

## Description
Fieldset used in terra-clinical-onset-picker does not contain a legend element. It is recommended that fieldsets should contain a legend element that provides a description of the field group.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-ui/#/components/terra-clinical-onset-picker/clinical-onset-picker/clinical-onset-picker
2. Inspect fieldset rendered by terra-clinical-onset-picker.

## Expected Behavior
Onset picker aligns with accessibility recommendations.

## Possible Solutions
1. Include legend with fieldset. This could include a prop that toggles if it is visually-hidden or not using terra-visually-hidden-text
2. Use a div as a wrapping container instead of a fieldset.

## Environment

* Component Name and Version: terra-clinical-onset-picker
* Browser Name and Version: All

Issue #435, opened 2018-12-20
Need ability to add legend for onset picker fieldsets


jmsv6d:
# Bug Report
While attempting to resolve accessibility issues for site examples for https://github.com/cerner/terra-ui/issues/46 the onset picker examples were not accessible as the fieldsets have no name associated to them available to accessibility APIs. Consumers should have the ability to add a legend for onset picker field sets for accessibility.

jmsv6d:
Duplicate #461

Issue #397, opened 2018-09-19
Violation in Onset Picker not resolving as expected


PaulSolDev:
# Bug Report

## Description

(Firefox only) When manually updating a value in the onset picker that causes a violation (exceeding the rules of 2 for Week >8 and Month >24) the corresponding field shows a violation.  Changing the unit value does not clear the violation.

Oddities- changing from units from weeks to months clears the violation as expected but changing from weeks to years or months to years does not.

## Steps to Reproduce

1. (in Firefox) Choose 'Age' in the Onset Picker
2. Select a unit of 'weeks'
3. Manually enter a value >8 and select an area outside of the field (violation appears)
4. Change the unit to years (note that the violation does not clear)
5. Select the value field and attempt to use the toggle up option (value does not change)
6. Select the toggle down option (unit drops to a significantly lower value and clears violation)
7. With a unit of years still selected change the value to >24 (notice that a violation occurs despite this being an accepted value for a unit of years)

## Additional Context / Screenshots

![image](https://user-images.githubusercontent.com/29487301/45774300-3803b380-bc12-11e8-9cd3-9a5f3f93e3f0.png)

![image](https://user-images.githubusercontent.com/29487301/45774282-30440f00-bc12-11e8-8b57-9320bff5334b.png)

## Expected Behavior

Follow rule of 2 and clear violations based on current field settings.

## Environment

* Component Name and Version: Onset Picker
* Browser Name and Version: Firefox
* Operating System and version (desktop or mobile): Desktop

ryanthemanuel:
@doogthemushroom is going to pick this up.

doogthemushroom:
> 1. (in Firefox) Choose 'Age' in the Onset Picker
> 2. Select a unit of 'weeks'
> 3. Manually enter a value >8 and select an area outside of the field (violation appears)

So this seems correct so far following rule of 2. This 'violation' is also happening on other browsers but none have this styling. If you followed these steps to reproduce on the handled picker example and clicked 'Submit' in between the steps, you'd see a popup occur in chrome at the very least.

> 4. Change the unit to years (note that the violation does not clear)

It's causing a violation here because the # of years puts the onset date in the future. In this case, the birthdate was 6 years ago and you are trying to set the date 8 years from the birthdate, that's a violation. No future dates.

> 5. Select the value field and attempt to use the toggle up option (value does not change)

Expected, as you shouldn't be able to enter a date in the future.

[...]

PR #468, opened 2019-02-22
Condense onChange handlers, add support for age units

doogthemushroom:
### Summary
Reducing the necessary number of onChange handlers from 3 to 1. Also adding a new `ageUnit` prop and onChange value to support initializing correctly with the `age` granularity. Updated some documentation to clarify the Onset Picker does not deal with time.

### Additional Details
To be clear this would be a breaking change, and would require a 4.0 release. I have added notes to the changelog regarding this.


packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -20,9 +20,15 @@ const GranularityOptions = {
...
 };
 
+const DATE_FORMAT = 'YYYY-MM-DD';
+
 const propTypes = {
   /**
-   * The ISO 8601 string representation of the birth date to calculate an onset date for the 'age' precision.
+   * The date unit of the age value. One of 'weeks', 'months', or 'years'.
+   */
+  ageUnit: PropTypes.string,
```
benbcai: Could use PropTypes.oneOf(['weeks', 'months', 'years']).

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -20,9 +20,15 @@ const GranularityOptions = {
   DATE: 'date',
 };
 
+const DATE_FORMAT = 'YYYY-MM-DD';
```
emilyrohrbough: shouldn't this format should be translated

doogthemushroom: No it's for internal use only, it's the date only portion of an ISO-8601 date.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
doogthemushroom: Updated to ['weeks', 'months', 'years']

tbiethman: I'd just go ahead and do `PropTypes.oneOf(['weeks', 'months', 'years'])`

doogthemushroom: Already updated per @benbcai suggestion!

tbiethman: D'oh i'm way out of date here

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -144,16 +151,14 @@ class OnsetPicker extends React.Component {
...
         };
-      }, () => this.handleDateChange(this.state.onsetDate));
-    }
-    this.setState({ granularity });
-
-    if (this.props.granularitySelectOnChange) {
-      this.props.granularitySelectOnChange(granularity);
+      }, () => this.handleOnsetUpdate());
+    } else {
+      this.setState({ granularity }, () => this.handleOnsetUpdate());
```
tbiethman: You could remove one function instantiation per execution if you provide handleOnsetUpdate directly as the callback: `this.setState({ granularity }, this.handleOnsetUpdate);`

doogthemushroom: Ah of course, updated. I think I like the look of these little anonymous functions too much.

emilyrohrbough:
Are any of these issues/bugs resolved with this pr? 
- https://github.com/cerner/terra-clinical/issues/397
- https://github.com/cerner/terra-clinical/issues/435
- https://github.com/cerner/terra-clinical/issues/461

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
emilyrohrbough: Can we add an inline comment to document this use?

PR #473, opened 2019-03-22
Remove terra-base peerDep and normalize react-intl version

bjankord:
### Summary
Work that resolves https://github.com/cerner/terra-core/issues/2291 for terra-clinical


packages/terra-clinical-application/src/Application.jsx
```diff
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import Base from 'terra-base';
```
bjankord: Looks like terra-base is a hard dependency for terra-clinical-application. I've left this is to keep the PR passive.

bjankord: Talked with the clinical team, plan to leave this import in here as is

PR #517, opened 2019-06-10
[terra-clinical-item-view] Updated misleading documentation

### LINKED ISSUE #513: [terra-clinical-item-view] Update misleading 'displays' documentation
# Bug Report
Update misleading displays documentation\

## Description
The displays proptype documentation is vague about the acceptable values elements for the component, as a result consumers mistakenly pass invalid elements. Should be updated to mention that in order to function properly displays should only be terra-clinical-item-display's.

https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-item-view/src/ItemView.jsx#L39

pranav300:
### Summary
Updated terra-clinical-item-view misleading `children` documentation.
Resolves #513

### DIFF
--- a/packages/terra-clinical-item-view/src/ItemView.jsx
+++ b/packages/terra-clinical-item-view/src/ItemView.jsx
@@ -36,7 +36,7 @@ const propTypes = {
    */
   endAccessory: PropTypes.node,
   /**
-   * An array of react display elements to be presented.
+   * An array of  terra-clinical-item-display's to be presented.
    */
   displays: PropTypes.arrayOf(PropTypes.element),
   /**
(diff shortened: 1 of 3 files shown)


packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -36,7 +36,7 @@ const propTypes = {
    */
   endAccessory: PropTypes.node,
   /**
-   * An array of react display elements to be presented.
+   * An array of terra-clinical-item-display's to be presented.
    */
   displays: PropTypes.arrayOf(PropTypes.element),
```
yuderekyu: if that's the case, should we scope prop type validation to the component too?
```suggestion
   displays: PropTypes.arrayOf(PropTypes.instanceOf(ItemDisplay),
```

PR #522, opened 2019-06-20
Resolve known WCAG 2.1 AA issues

bjankord:
### Summary
Resolve known WCAG 2.1 AA issues


packages/terra-clinical-data-grid/src/DataGrid.jsx
```diff
@@ -1051,7 +1051,7 @@ class DataGrid extends React.Component {
       >
         <div
           role="button"
-          aria-hidden
+          aria-label="Navigate to date grid"
```
bjankord: Need to get translations for these strings.

jmsv6d:
```suggestion
          aria-label="Navigate to data grid"
```

bjankord: Translations should be coming this Friday

PR #528, opened 2019-07-03
Removes legacy react contexttypes and uses formatmessage from react-intl.

lokesh-0813:
### Summary
Few packages which were interfacing with legacy react `contexttypes` are removed. Now the packages use an abstraction like `formatMessage` to interface with `injectIntl`.

Solves : #523


packages/terra-clinical-onset-picker/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Removed react legacy `contexttypes`.
+* Updated the component with `formatMessage` to interface with `react-intl`.
```
nramamurth: I think you meant to say `injectIntl` here.

lokesh-0813: I will have it updated. Thank you.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -85,6 +86,11 @@ const propTypes = {
    * Whether or not the legend is visible. Use this props to hide a legend while still creating it on the DOM for accessibility.
    */
   isLegendHidden: PropTypes.bool,
+
+  /**
+   * The intl object containing translations. This is retrieved from the context automatically by injectIntl.
```
bjankord: We should add the @private syntax so this doesn't show up in the generated prop-types documentation.
```suggestion
   * @private
   * The intl object containing translations. This is retrieved from the context automatically by injectIntl.
```

bjankord:
I'm talking with the team about this update and will post back once I have some consensus on the direction we want to go with this.

bjankord:
@lokesh-0813 Talked about this with @ryanthemanuel. The plan is to move forward with `injectIntl` in the places that you've identified where it doesn't make sense to use `<FormattedMessage />`

lokesh-0813:
Sticking to `injectIntl` since `formattedMessage` does not make sense.

PR #529, opened 2019-07-08
[terra-clinical-header] Update styles

yuderekyu:
### Summary
Update styles to better align with action-header. Add two theme variables to control height.


packages/terra-clinical-header/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Unreleased
 
 ### Changed
```
StephenEsser: This isn't a result of this pull request, but here are two unique changed sections now under unreleased. Would you mind cleaning them up?

packages/terra-clinical-header/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Unreleased
...
 * Update wdio spec to use Terra.describeViewports
+* Remove absolute positioning from title container
+* Add new theme variables for height
+** --terra-clinical-header-height
+** --terra-clinical-header-subheader-height
```
StephenEsser: These changes might make more sense under an "Added" section. Which ever section they end up in they might need a different syntax:

image

yuderekyu: updated https://github.com/cerner/terra-clinical/pull/529/commits/9bccacf8d5d239e892137d748dfbd6ec47a0039f

packages/terra-clinical-header/CHANGELOG.md
yuderekyu: updated https://github.com/cerner/terra-clinical/pull/529/commits/3977fc628ed9333cc59b80ad72038ce2c500b58d

packages/terra-clinical-header/CHANGELOG.md
```diff
@@ -3,12 +3,18 @@ ChangeLog
...
 ----------
-### Changed 
-* Replaced inline stylings for terra-dev-site components with .scss files
-
 ### Changed
+* Replaced inline stylings for terra-dev-site components with .scss files
 * Update wdio spec to use Terra.describeViewports
 
+### Added
+* Add new theme variables for height
```
mjhenkes: Should we do a pr to the dark theme and consumer theme repo before we add these?

yuderekyu: I'll open one for consumer. I can't speak for radiology, but judging from the pull request @bjankord opened in their space, they'd prefer to handle changes themselves.

@cwalten @PseudoFaux should I create a pull request?

yuderekyu: Actually, consumer doesn't theme any clinical components

mjhenkes: open one up for radiology, we mis-understood their pr process and they accepted the change.

mjhenkes: On review of the themes, it seems like we're fine to go forward without prs

yuderekyu: The theme does not define theme variables for css custom attributes that use default values. We'll let radiology handle this change if they want a different height.

PR #532, opened 2019-07-16
Add ability to set height on specific rows

### LINKED ISSUE #530: [terra-clinical-data-grid] Add ability to set individual row heights
# Feature Request

## Description
Investigate and add Exposing ability to custom set specific heights per row on a selective basis. Current API would persist of setting single row height for all rows unless specified with new API. 

(Additional details may be added)

## @ Mentions
@tbiethman

Blackop778:
### Summary
Individual rows can have a height set. If a height is not set on a row then the `DataGrid`'s `rowHeight` property will be applied.

### Additional Details
Closes #530


packages/terra-clinical-data-grid/docs/README.md
```diff
@@ -59,10 +59,11 @@ Rows are provided to sections to be rendered by the DataGrid. The order in which
...
 |---|---|---|---|---|
 |**id**|string|required|none|An identifier for the row. This identifier should be unique across the set of rows provided to a given section.|
-|**ariaLabel**|string|optional|A string identifier used to describe the row contents. This value will be used to contruct additional labels for internal controls (e.g. row selection cells).|
+|**ariaLabel**|string|optional|none|A string identifier used to describe the row contents. This value will be used to contruct additional labels for internal controls (e.g. row selection cells).|
 |**isSelectable**|bool|optional|none|A boolean indicating whether or not the row is selectable.|
 |**isSelected**|bool|optional|none|A boolean indicating whether or not the row should render as selected.|
 |**cells**|array|optional|[]|An array of cell objects that define the content to be rendered in the row. The order in which the cells are rendered is determined by the order of the `pinnedColumns` and `overflowColumns` props.|
+|**height**|string|optional|none|The height for this row. This overrides the `DataGrid` `rowHeight` property. Any valid CSS height value is accepted (i.e. 50px, 3rem, etc.)|
```
tbiethman: In light of new better-defined accessibility requirements, we should call out `rem` units as the preferred unit.

(If there are any px's hanging out in the examples, we could update those too).

Blackop778: Done in a7e965333ae07c990b02a5592a5a83e0bd0bcdda and 4e29385d3a5e8e36195ea8e9ecf2179b984b874a

packages/terra-clinical-data-grid/src/DataGrid.jsx
```diff
@@ -60,11 +60,12 @@ const propTypes = {
    */
   onRequestSectionCollapse: PropTypes.func,
   /**
-   * String that specifies the row height. Any valid CSS height value is accepted (i.e. 50px, 3rem, etc.)
+   * String that specifies the row height. Values are suggested to be in `rem`s (ex `'5rem'`), but any valid CSS height value is accepted.
+   * This value can be overridden by for a row by specifying a height on the given height.
```
nramamurth: This should be rephrased. I think this was meant to read, “This value can be overridden by a height that is exclusively specified for a given row”?

Blackop778: Good catch, fixed in cde2bd7c871a86afc4a36131a6f0bc94fa52cd5a

PR #541, opened 2019-07-25
[clinical-onset-picker] Blank Age fixed to return blank onset date.

### LINKED ISSUE #536: Inconsistent "blank" values for Onset Picker
# Bug Report

## Description
Onset picker is handling blank values inconsistently.  A blank "Date" granularity returns an onsetDate of "".  A blank "Age" granularity returns the birth date.

## Steps to Reproduce

Reproduced on the terra example pages. Example: "Onset picker with supplied data and handlers"
1. Change granularity to "Age"
2. Clear the number
3. Hit Submit - note that the onset date is the birthdate

4. Change granularity to "Date"
5. Clear the date
6. Hit Submit - note that the onset date is blank.

## Expected Behavior
If an onset age is not set I would expect the value to be blank as well.

## Possible Solution
Age is always getting set to a number [here](https://github.com/cerner/terra-clinical/blob/ca3e23209461bba0001db7fd4d388497bbe71dfe/packages/terra-clinical-onset-picker/src/OnsetPicker.jsx#L179).  If the inputs value is not set, clear out data instead of converting it to 0.

## Environment

* Component Name and Version:
* Node/npm Version: Node 8/npm 5
* Webpack Version: 4.34.0
* Operating System and version (desktop or mobile): All

## @ Mentions
@doogthemushroom, @ryanthemanuel

saket2403:
### Summary
Resolves #536

### DIFF
--- a/packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
+++ b/packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
@@ -153,7 +153,7 @@ class OnsetPicker extends React.Component {
           granularity,
           age: ageValues.age,
           ageUnit: ageValues.ageUnit,
-          onsetDate: moment(this.props.birthdate).add(ageValues.age, ageValues.ageUnit),
+          onsetDate: undefined,
         };
       }, this.handleOnsetUpdate);
     } else {
@@ -177,19 +177,25 @@ class OnsetPicker extends React.Component {
    */
   changeAge(event) {
     const age = Number(event.target.value);
-
-    this.setState((prevState) => {
+    if (event.target.value !== '') {
+      this.setState((prevState) => {
       // Check if date can be calculated
-      const ageDate = Number.isInteger(age) && prevState.ageUnit
-        ? moment(this.props.birthdate).add(age, prevState.ageUnit) : undefined;
-      // Check if date is valid
-      const validDate = ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment();
+        const ageDate = Number.isInteger(age) && prevState.ageUnit
+          ? moment(this.props.birthdate).add(age, prevState.ageUnit) : undefined;
+        // Check if date is valid
+        const validDate = ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment();
 
-      return {
+        return {
+          age,
+          onsetDate: validDate ? ageDate : undefined,
+        };
+      }, this.handleOnsetUpdate);
+    } else {
+      this.setState(() => ({
         age,
-        onsetDate: validDate ? ageDate : undefined,
-      };
-    }, this.handleOnsetUpdate);
+        onsetDate: undefined,
+      }), this.handleOnsetUpdate);
+    }
   }
 
   /**
@@ -198,18 +204,25 @@ class OnsetPicker extends React.Component {
    * @param {ageUnit} - New ageUnit
    */
   changeAgeUnit(ageUnit) {
-    this.setState((prevState) => {
+    if (this.state.age !== undefined) {
+      this.setState((prevState) => {
       // Check if date can be calculated
-      const ageDate = Number.isInteger(prevState.age) && ageUnit
-        ? moment(this.props.birthdate).add(prevState.age, ageUnit) : undefined;
-      // Check if date is valid
-      const validDate = ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment();
+        const ageDate = Number.isInteger(prevState.age) && ageUnit
+          ? moment(this.props.birthdate).add(prevState.age, ageUnit) : undefined;
+        // Check if date is valid
+        const validDate = ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment();
 
-      return {
+        return {
+          ageUnit,
+          onsetDate: validDate ? ageDate : undefined,
+        };
+      }, this.handleOnsetUpdate);
+    } else {
+      this.setState(() => ({
         ageUnit,
-        onsetDate: validDate ? ageDate : undefined,
-      };
-    }, this.handleOnsetUpdate);
+        onsetDate: undefined,
+      }), this.handleOnsetUpdate);
+    }
   }
 
   /**
--- a/packages/terra-clinical-onset-picker/CHANGELOG.md
+++ b/packages/terra-clinical-onset-picker/CHANGELOG.md
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Blank Age granularity fixed to return a blank onset date value.
 
 4.4.0 - (July 16, 2019)
 ----------


pranav300:
When Initially using granularity as age and ageUnits as years. I am seeing that onsetDate is not getting updated.
chrome-capture

And for the same inputs in the mobile view, the result is overflowing.
chrome-capture

emilyrohrbough:
Can you add test cases to validate this change?

saket2403:
> And for the same inputs in the mobile view, results are overflowing.

So I have figured out 2 fixes- 
1. To use 'break-word' which might break the overflowing word arbitrarily depending on resolution. 
2. To display  2 parameters each in separate lines.

Which of these should I go through with?

2 lines


Word Break

neilpfeiffer:
@saket2403 Since the doc-site example just needs something basic to return the displays, all you need to do is add some space as the last [stringify parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) and the browser will naturally break-word without any extra work:

For [.../terra-dev-site/doc/example/HandledOnset.jsx#L68](https://github.com/cerner/terra-clinical/blob/ef9e69746f798fad7798fbf48c51eaaeb01c54df/packages/terra-clinical-onset-picker/src/terra-dev-site/doc/example/HandledOnset.jsx#L68) -- change to:
            `<p>{JSON.stringify(this.state.submittedData, null, 2)}</p>`

This will affect screenshots, so they will need to be re-run.

----

Still needing to be resolved are:
> When Initially using granularity as age and ageUnits as years. I am seeing that onsetDate is not getting updated.
(this is not broken in the current code, 
and 

> Can you add test cases to validate this change?

I would make a new wdio test that copies this one: [Terra.it.validatesElement('age input changed');](https://github.com/cerner/terra-clinical/blob/ef9e69746f798fad7798fbf48c51eaaeb01c54df/packages/terra-clinical-onset-picker/tests/wdio/onset-spec.js#L234-L240) but adds the step to focus in the input and clear the value. Test could be `Terra.it.validatesElement('age input cleared');`

saket2403:
All comments resolved in 60fc9e6 and 55b3168

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -174,20 +174,27 @@ class OnsetPicker extends React.Component {
    * @param {event} - Triggered change event
    */
   changeAge(event) {
-    const age = Number(event.target.value);
-
-    this.setState((prevState) => {
+    if (event.target.value !== '') {
+      const age = Number(event.target.value);
```
ShettyAkarsh: we can declare this once at the beginning use LET and can be reused in else block also.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -174,20 +174,27 @@ class OnsetPicker extends React.Component {
    * @param {event} - Triggered change event
    */
   changeAge(event) {
-    const age = Number(event.target.value);
-
-    this.setState((prevState) => {
+    if (event.target.value !== '') {
```
ShettyAkarsh: can we check this using 'Undefined' or null ? is this value be always be ' ' ? if yes, then declare it as const such as

const BlankValue = ' '; in the beginning of the file. so that it can be reused across the class.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
supreethmr: in else block you can remove age : [here](https://github.com/cerner/terra-clinical/blob/5d0df5ef328f6245d74e7c707014a9ebb166f02e/packages/terra-clinical-onset-picker/src/OnsetPicker.jsx#L192) and can set undefined [here](https://github.com/cerner/terra-clinical/blob/5d0df5ef328f6245d74e7c707014a9ebb166f02e/packages/terra-clinical-onset-picker/src/OnsetPicker.jsx#L194)

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -196,18 +203,25 @@ class OnsetPicker extends React.Component {
...
+      this.setState((prevState) => {
       // Check if date can be calculated
-      const ageDate = Number.isInteger(prevState.age) && ageUnit
-        ? moment(this.props.birthdate).add(prevState.age, ageUnit) : undefined;
-      // Check if date is valid
-      const validDate = ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment();
+        const ageDate = Number.isInteger(prevState.age) && ageUnit
+          ? moment(this.props.birthdate).add(prevState.age, ageUnit) : undefined;
+        // Check if date is valid
+        const validDate = ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment();
```
ShettyAkarsh: so instead of if, else block in this method, can we do this ?

 const validDate = (this.state.age !== undefined) ? (ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment()) : undefined

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
tbiethman: The empty string is false-y anyway, so you should be able to cover all cases with `if (event.target.value) {}`

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Resolved in 7afc761

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Resolved in 7afc761

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Since this.state.age is the same as prevState.age  leaving this line as it is.

ShettyAkarsh: in that case can we use like this by declaring one more variable.

const age = (prevState.age !== undefined) ? Number.isInteger(prevState.age) : undefined
const ageDate = age && ageUnit ? moment(this.props.birthdate).add(prevState.age, ageUnit) : undefined;
const validDate = (this.state.age !== undefined) ? (ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment()) : undefined

saket2403: Yes this works fine . Changed in 8461b22

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -196,18 +203,15 @@ class OnsetPicker extends React.Component {
...
-        ? moment(this.props.birthdate).add(prevState.age, ageUnit) : undefined;
-      // Check if date is valid
-      const validDate = ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment();
-
-      return {
-        ageUnit,
-        onsetDate: validDate ? ageDate : undefined,
-      };
-    }, this.handleOnsetUpdate);
+    const age = (this.state.age !== undefined) ? Number.isInteger(this.state.age) : undefined;
```
doogthemushroom: `Number.isInteger` will return `false` on `undefined`, so I wouldn't bother with the check. Also wouldn't name this `age` as it's a boolean value, and not actually the age value. I'd also probably not set these values to `undefined`, when you're using them as booleans. Yes it's a falsey value, but why not just use `false`? Suggestion:

```javascript
const validAge = Number.isInteger(this.state.age);
// Attempt to calculate onset date
const ageDate = validAge && ageUnit ? moment(this.props.birthdate).add(this.state.age, ageUnit) : undefined;
// Check if date is valid
const validDate = validAge ? (ageDate && ageDate >= moment(this.props.birthdate) && ageDate <= moment()) : false;
```

saket2403: Resolved in 982e7aa

PR #543, opened 2019-07-25
[terra-clinical-onset-picker] crashing for age less than a week.

### LINKED ISSUE #535: Onset Picker crashes when birthdate is less than one week ago
# Bug Report

## Description
When a birthdate prop is set that is less than 1 week ago, the onset control crashes when granularity is set to age.

Error is thrown on this [line](https://github.com/cerner/terra-clinical/blob/ca3e23209461bba0001db7fd4d388497bbe71dfe/packages/terra-clinical-onset-picker/src/OnsetPicker.jsx#L378-L379) because allowedAgeUnits is returning null.

## Steps to Reproduce
1. Set the onset control birthdate prop to less than one week ago.
2. Set the granularity of the control to age.

## Expected Behavior
Control does not crash.

## Possible Solution
The age granularity may not make sense for patients that are less than a week old.  Allowed age units are returned [here](https://github.com/cerner/terra-clinical/blob/ca3e23209461bba0001db7fd4d388497bbe71dfe/packages/terra-clinical-onset-picker/src/OnsetUtils.js#L92-L94) and null is returned.

## Environment

* Component Name and Version: terra-clinical-onset-picker/4.3.0
* Node/npm Version: Node 8/npm 5
* Webpack Version: 4.34.0
* Operating System and version (desktop or mobile): desktop

## @ Mentions
@doogthemushroom, @ryanthemanuel

pranav300:
### Summary
I have removed the age granularity option whenever age is less than a week.
Resolves #535


packages/terra-clinical-onset-picker/src/terra-dev-site/test/clinical-onset-picker/LessWeek.test.jsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import moment from 'moment';
+import OnsetPicker from '../../../OnsetPicker';
+
+class HandledOnsetExample extends React.Component {
```
supreethmr: nit: class name can be changed to LessWeek or something more suitable for this test.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -294,7 +294,16 @@ class OnsetPicker extends React.Component {
     const { intl } = this.context;
 
     let granularitySelect = null;
+    const birthMoment = moment(birthdate).startOf('day'); // startOf to clear time from values
+    const currentMoment = moment().startOf('day');
     if (this.state.precision !== PrecisionOptions.UNKNOWN) {
+      const optionAge = (currentMoment.diff(birthMoment, 'weeks') !== 0) ? (
```
supreethmr: Since condition for age less than a week is being handled here now, do we still need to the same check in [allowedAgeUnits](https://github.com/cerner/terra-clinical/blob/8156586a2a2eb261d4f732c19f76a5ec0b52d39c/packages/terra-clinical-onset-picker/src/OnsetUtils.js#L92) method of OnsetUtils.js

packages/terra-clinical-onset-picker/src/terra-dev-site/test/clinical-onset-picker/LessWeek.test.jsx
```diff
@@ -0,0 +1,68 @@
...
+      },
+    }));
+  }
+
+  render() {
+    const birthdate = moment().subtract(6, 'days');
+    return (
+      <div>
+        <p>Onset picker with age to be less than a week.</p>
+        <p>{`Current date locked to ${moment().format('MM-DD-YYYY')}`}</p>
```
supreethmr: To make the purpose of this test more clear I feel adding birthdate also here would be appropriate.

pranav300: Fixed here https://github.com/cerner/terra-clinical/pull/543/commits/dea7f66925ff7547256295d24794eb437c9dc8ca.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
pranav300: Fixed here https://github.com/cerner/terra-clinical/pull/543/commits/dea7f66925ff7547256295d24794eb437c9dc8ca.

packages/terra-clinical-onset-picker/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Unreleased
 4.5.0 - (August 14, 2019)
 ----------
 ### Changed
+* Removed age granularity from being populated for age less than a week.
```
StephenEsser: This must have moved into a released section while merging maser, we'll need to add it under unreleased. We'll want to place it within a `# Fixed` section as well

pranav300: Fixed here https://github.com/cerner/terra-clinical/pull/543/commits/37b811e87bebdcdb25bd349920472977e3217844.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -292,7 +292,16 @@ class OnsetPicker extends React.Component {
...
+        <SelectField.Option
+          value={GranularityOptions.AGE}
+          display={intl.formatMessage({ id: 'Terra.onsetPicker.age' })}
+          key={GranularityOptions.AGE}
+        />
+      ) : <React.Fragment />;
```
StephenEsser: It's a little strange to return a raw fragment here, I think you can return undefined.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -309,11 +318,7 @@ class OnsetPicker extends React.Component {
           }}
           selectId={`${this.props.id}-granularity-select`}
         >
-          <SelectField.Option
-            value={GranularityOptions.AGE}
-            display={intl.formatMessage({ id: 'Terra.onsetPicker.age' })}
-            key={GranularityOptions.AGE}
-          />
```
StephenEsser: This is probably a question for UX / SD's. Having the option disappear completely based on age might be confusing to users. 

A couple potential alternatives:

1. Disabled the age option if the age falls below 1 week. This preserves the option within the list, but makes it un-selectable. ( This would need UX / SD approval, disabled options are sometimes frowned upon )
2. Allow the user to select an age granularity but display a new message or component if the age falls below 1 week. ( instead of an empty Select )

cc: @neilpfeiffer @PaulSolDev

packages/terra-clinical-onset-picker/src/OnsetUtils.js
```diff
@@ -89,10 +89,6 @@ class OnsetUtils {
     const ageMoment = moment(birthdate).startOf('day'); // startOf to clear time from values
     const currentMoment = moment().startOf('day');
 
-    if (currentMoment.diff(ageMoment, 'weeks') === 0) {
-      return null;
```
StephenEsser: We should probably preserve returning null or return an empty array here. Otherwise a call to `allowedAgeUnits` can / will return age units that are not actually allowed.

doogthemushroom: Theoretically this should never get called with an age less than a week old since we're removing the option in the OnsetPicker render. So whenever the age option is available, the week option would minimally always be there. That said, this didn't need to be removed either and does reinforce the correct "Allowed Age Units".

doogthemushroom:
This doesn't seem to be handling the scenario where the consumer initializes the OnsetPicker with age granularity and a birthdate of less than a week. With the PR code I would assume that would still render the age select with the 'week' option allowed but have a blank granularity. That has a couple of issues. I believe we should be changing the granularity to the default value in this scenario.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
pranav300: So, when I put `undefined` instead of a `raw fragment`. At that time the code breaks at this line [SelectUtils]( https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/shared/_SelectUtil.js#L101)

packages/terra-clinical-onset-picker/src/OnsetUtils.js
pranav300: updated here https://github.com/cerner/terra-clinical/pull/543/commits/cbaf09e53203239472adb85c23b4a658a641ae83

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
StephenEsser: The logic may need to change to either conditionally push these into an array or conditionally add it below. I don't think an empty fragment should be added as a workaround, the option should be removed from the list instead of substituted.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -128,8 +128,14 @@ class OnsetPicker extends React.Component {
       ageValues = OnsetUtils.onsetToAge(this.props.birthdate, moment(this.props.onsetDate));
     }
 
+    const birthMoment = moment(props.birthdate).startOf('day');
+    const currentMoment = moment().startOf('day');
+    /* The granularity is added so that if 'age' is passed as granularity with invalid birthdate,
+    then the qranularity is set to default i.e 'undefined' */
```
doogthemushroom: qranularity

pranav300: oops

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
pranav300: updated here https://github.com/cerner/terra-clinical/pull/543/commits/0c708f260e59a74b02bcb9a9a6e76e5a2035e4c3.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -294,9 +300,20 @@ class OnsetPicker extends React.Component {
       ...customProps
     } = this.props;
 
-
     let granularitySelect = null;
+    const optionAge = [];
+    const birthMoment = moment(birthdate).startOf('day'); // startOf to clear time from values
+    const currentMoment = moment().startOf('day');
```
benbcai: The declaration of birthMoment and currentMoment can be moved inside the `if` block below as they are not needed for the UNKNOWN precision.

pranav300: fixed here https://github.com/cerner/terra-clinical/pull/543/commits/5ac6141da4edeabdf510c7664b89ade9d91e4637.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
jeremyfuksa: **UX Input**

My design standards discussion ruled out option 1 since our design philosophy on disabled items is that the assumption is that there's something that can be done elsewhere to enable that item. So, since birthdate is fixed, that can't be.

Option 2 adds UI cruft.

Option 3 is the best option. But, also according to design standards, we should contextually only give options that are applicable. So, in this case, until a patient is 1 week old, age should disappear. And, perhaps Year would disappear until the patient is one year old. Month could still be applicable for that age patient, but Date seems like a default choice for a patient less than one week old.

jeremyfuksa: However, there's more at play here that could inform other design solutions. [I replied in the original issue](https://github.com/cerner/terra-clinical/issues/535#issuecomment-540799446) in order to get more context. If this is a NICU scenario, it makes sense to have granularity down to hours or even minutes.

ryanthemanuel: @jeremyfuksa for the current consumers of this control, hours and minutes aren't supported as a mechanism to pick the age.  They actually get stored in the database as a number and a unit and the only units available are days, weeks, months.  If we have a consumer in the future that supports hours and minutes, I think we could extend the control to allow those additional units and update the logic to drop age if the patient is less than the minimum amount old, but for now I think we just drop the age if they're under a year.

PaulSolDev:

PR #562, opened 2019-08-27
[detail-view, label-value-view] Removed item-display as a dependency.

### LINKED ISSUE #561: Remove problematic terra-clinical-item-display as a dependency
# Bug Report
Remove terra-clinical-item-display from examples in components that have no functional need for it's consumption.

## Description
Using ItemDisplays outside of the sanctioned use-case has caused confusion and mistakes within our consumer's implementations. The component should only be assocaited with the terra-clinical-item-view and/or terra-clinical-item-collection.

Components to be updated:
terra-clinical-label-value-view
terra-clinical-detail-view

## @ Mentions
@mjhenkes

pranav300:
### Summary
Resolves #561


packages/terra-clinical-label-value-view/src/terra-dev-site/doc/example/LabelValueView.module.scss
```diff
@@ -5,4 +5,23 @@
...
+
+  .icon {
+    flex: 0 0 auto;
+    overflow: hidden;
+    padding-right: 0.357142857142857rem;
+  }
+
+  .text {
+    color: #c00;
+    font-weight: 700;
```
avinashg1994: When we specify the "textStyle" as "attention", the styles of both the classes, i.e. `text` and `attention`, are applied.
So, you have to also add these styles : https://github.com/cerner/terra-clinical/pull/562/files#diff-ea4d170f02cea60d2157a80140ea9d5fR21-R22.
```suggestion
    font-weight: 700;
    flex: 1 1 auto;
    min-width: 10px; // Need min-width, as IE10 will completely squish display
```

PR #565, opened 2019-09-12
[onset-picker] Unknown precision fixed to return correct values.

### LINKED ISSUE #542: Wrong values returned for unknown precision in Onset Picker
# Bug Report

## Description
Onset Picker is returning the previously set values for precision unknown. 

## Steps to Reproduce
Reproduced on the terra example pages. Example: "Onset picker with supplied data and handlers"

1.Select any precision except unknown
2. Select a granularity.
3. Change precision to unknown.
4. Hit Submit

## Current Behaviour
Returns the previous values with only precision set to unknown.

## Expected Behaviour
Output should be - {"precision":"unknown","onsetDate":""}

## Environment
* Component Name and Version: Onset Picker
* Node/npm Version:  Node 8
* Webpack Version: 4.36.1
* Operating System and version (desktop or mobile): All

## @ Mentions
@ryanthemanuel  @doogthemushroom

saket2403:
Fixes #542


packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -268,11 +275,10 @@ class OnsetPicker extends React.Component {
 
     const onsetObject = {
       precision: this.state.precision,
-      granularity: this.state.granularity,
       onsetDate: this.state.onsetDate ? this.state.onsetDate.format(DATE_FORMAT) : '',
+      granularity: this.state.precision !== PrecisionOptions.UNKNOWN ? this.state.granularity : '',
```
StephenEsser: Granularity should always be undefined for Unknown at this point in the code. Is there ever a case during this check where this isn't true? 

This check should be unnecessary. If for passivity we want to return a string instead of undefined we can `or` the assignment. 

```jsx
granularity: this.state.granularity || '', // Granularity should always be undefined for Unknown precision
```

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -165,7 +165,14 @@ class OnsetPicker extends React.Component {
    * @param {precision} - New precision value
    */
   changePrecision(precision) {
-    this.setState({ precision }, this.handleOnsetUpdate);
+    if (precision !== PrecisionOptions.UNKNOWN) {
```
StephenEsser: This is up to your preference, but when possible I think taking the truthy path is usually easier to understand. I'd recommend flipping this statement so we always perform an action when it is Unknown rather than if it's not.

```jsx
if (precision === PrecisionOptions.UNKNOWN) {
  // Logic for Unknown
} else {
  // Logic for any other precision
}
```

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
doogthemushroom: When I first created the component I just didn't clear the non-precision values out when setting precision to Unknown. My assumption was that if you get an Unknown precision the other values are irrelevant. But it should be fine and is probably a good idea to just clear out non-precision values.

ryanthemanuel: in talking with @doogthemushroom he thinks we should be clearing the values on change.  it may have been something that was missed.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Resolved in 6c98210

PR #569, opened 2019-09-13
Pin webpack version to 4.39.3

bjankord:
### Summary
Temporary resolution to fix issue related to buggy behavior with webpack v4.40.0

Related:
* https://github.com/webpack/webpack/issues/9692
* https://github.com/webpack/webpack/issues/9693


bjankord:
These seems to be no longer needed with the 4.40.1 release of webpack.

PR #596, opened 2019-12-12
[terra-clinical-onset-picker] Added css to align age-field correctly in IE.

### LINKED ISSUE #586: [terra-clinical-onset-picker] age field not aligned properly in IE.
# Bug Report

## Description
The age field is getting misaligned in IE.

## Steps to Reproduce

1. Go to [clinical-onset-picker](https://engineering.cerner.com/terra-ui/components/terra-clinical-onset-picker/clinical-onset-picker/clinical-onset-picker)
2. Go to any of the examples. And change granularity to age

## Additional Context / Screenshots

VirtualBox_Windows 7_08_11_2019_10_53_23

## Expected Behavior

The age field must be aligned correctly in IE.

## Possible Solution

Add CSS property `display: flex;` to [this](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-onset-picker/src/OnsetPicker.jsx#L495) div.

## Environment

* Component Name and Version: clinical-onset-picker v4.10.0
* Browser Name and Version: IE11
* Operating System and version (desktop): Windows 10

pranav300:
### Summary
Fixes #586


emilyrohrbough:
Are there screenshots that capture this that need updated?

packages/terra-clinical-onset-picker/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ ChangeLog
 Unreleased
 ----------
 
+### Added
```
emilyrohrbough: Lets change this to `Fixed` since its a bug fix

pranav300:
> Are there screenshots that capture this that need updated?

I have added a media query specifically for IE [here](https://github.com/cerner/terra-clinical/pull/596/commits/f5d6ee66c7210db488270e9b123d0c77d739885b), as other browsers shouldn't be affected by this change. 
What do you think? @emilyrohrbough

packages/terra-clinical-onset-picker/src/OnsetPicker.module.scss
```diff
@@ -44,4 +44,10 @@
   .year {
     min-width: var(--terra-clinical-onset-picker-year-select-min-width, 5.5em);
   }
+
+  .content-wrapper {
+    @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) { // -ms needed for icon alignment in IE
+      display: flex;
```
StephenEsser: Do we know what is causing the misalignment? Applying a flex display style for only IE browsers seems like it might be masking a different issue. This solution does work, but it would be ideal if we knew what was is introducing the misalignment.

pranav300:
>If we remove the media query, does the picker align incorrectly on non-IE browsers?

Yes. So, it does misalign on IE and non-IE browsers when the media query does not have the breakpoint constraint - I noticed that media query sans the breakpoint constraint breaks the age field at `tiny` viewport regardless of the browser.

An alternative that I can think of is to have `float: left` [on the age field](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-onset-picker/src/OnsetPicker.module.scss#L5) and not go via the media query route.

ryanthemanuel:
@pranav300 let's investigate removing the media query if we can.

packages/terra-clinical-onset-picker/src/OnsetPicker.module.scss
jeremyfuksa: I've messed around with every box model tweak I can think of, and other than noticing that the misalignment is exactly 1rem, I can't figure out why this is initially misaligned.

emilyrohrbough: @jeremyfuksa could the issue be the default vertical-align value is baseline? I found some articles that might give more info:
- https://stackoverflow.com/questions/13548168/why-my-inline-block-divs-are-not-aligned-when-only-one-of-them-has-text
- https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align

dkasper-was-taken:
You might look at a more holistic approach to the solution rather than something dependent on media query. Here is an example:
https://github.com/cerner/terra-clinical/compare/proposed-onset-changes

packages/terra-clinical-onset-picker/src/OnsetPicker.module.scss
```diff
@@ -8,6 +8,7 @@
 
     .field-inline {
       display: inline-block;
+      vertical-align: top;
```
neilpfeiffer: After a bit of investigation -- @jeremyfuksa AND @emilyrohrbough were [right on the nose](https://github.com/cerner/terra-clinical/pull/596#discussion_r364337689). All we needed to do was fix `vertical-align` on the field component. 

I couldn't figure out why it wasn't already coming through directly to the Onset-picker, since Onset-picker uses the Terra-Field component and this is already fixed for IE, but then I found that Onset-picker adds it's own className for `field-inline` to each form field, and does not use the `isInline` prop as I was expecting (probably to avoid needing to override the extra spacing, might be my guess). Example here:
https://github.com/cerner/terra-clinical/blob/44bc65409b07fe940c9d0bb5add6bd45d96ce4d5/packages/terra-clinical-onset-picker/src/OnsetPicker.jsx#L373-L376

For a short-term & quick fix, I have added the `vertical-align: top` to the `.field-inline` class here in onset picker css to match the same styling that Terra-Field has in it's `.field-inline` class here: 
[../Field.module.scss#L9-L13](https://github.com/cerner/terra-core/blob/bc075f286cd9a548094a79f2f721ec526e2b251a/packages/terra-form-field/src/Field.module.scss#L9-L13), and removes the need for any additional media queries, etc.

However, it might be worthwhile to revisit and test swapping out the internal class to the actual isInline prop with any appropriate spacing overrides as needed, with some extensive cross-browser testing to make sure it doesn't introduce any other layout bugs. Can be discussed later.

PR #597, opened 2019-12-13
[terra-onset-picker] Clear existing data when granularity changes.

### LINKED ISSUE #593: [clinical-onset-date-picker] Onset Picker Doesn't Calculate Age Correctly When Given Leap Day Birthdate
# Bug Report

## Description

When a leap day birthdate is given to the onset picker (e.g. '2000-02-29' which is a leap day) when changing granularities it miscalculates the age.

## Steps to Reproduce

1. Give the onset picker a leap day birthdate (e.g. '2000-02-29')
2. Switch granularity to year and enter 2003 (a non-leap-year)
3. Switch granularity to age and witness that the correct age of 3 is not shown but 2.

## Expected Behavior

I expect the onset picker to correctly calculate leap-day birthdate ages.

## Possible Solution

If given a leap day birthdate, it should fall forward a day and default to 'YYYY-03-01' or March 01 YYYY to calculate the age field.

## Environment

* Component Name and Version: "terra-clinical-onset-picker": "4.10.0"
* Browser Name and Version: Chrome Version 78.0.3904.108
* Node/npm Version: [e.g. Node 8/npm 5] v8.9.2/5.5.1
* Webpack Version: 4.41.2
* Operating System and version (desktop or mobile): macOS Mojave 10.14.6

nramamurth:
### Summary
Fixes #593. 
### New Requirements
If Initial granularity and onset do not meet expected logic then the granularity should instead default to 'date' and the onset date is shown.
Note: any supplied precision value is maintained.

**Expected logic:**
- Reset all data except Precision when user changes the Granularity.
- Granularity: year --then month and day should always be January 1st
- Granularity: month --then the day should always be the 1st of the month
- Granularity: age, Age unit: years --then the month and day should always match the supplied birthdate month and day (*exception only when a leap year birthdate scenario exists. In the case with a leap year birthdate to a non leap year onset the day would advance to March 1st )
- Granularity: age, Age unit: months --then the day should always match the supplied birthdate day (**exception when the birthdate day of month is greater than the available number of days in the onset month.  Example: birthday October 31st -- for months with <31 days the system should calculate the onset to the 1st day of the next month)
- Granularity: age, Age unit: weeks --then the day should always match the calculated number of days forward from the birthdate


packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -348,6 +452,7 @@ class OnsetPicker extends React.Component {
 
     let ageInput = null;
     let ageUnitSelect = null;
+    let yearSelect = null;
```
yuderekyu: Can we declare these three variables right before each conditional setter? I think this is easier to follow. Also, nitpick, but we could leave off the null assignment - React will ignore both undefined and null children.
```jsx
let yearSelect;
if (this.state.granularity === GranularityOptions.YEAR) {
  yearSelect = this.getYearInput(intl, this.props.id);
}
```

packages/terra-clinical-onset-picker/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Clear existing data when granularity is changed.
```
nramamurth: It'd be great if you could briefly summarize all the changes that have been made as part of this PR in the PR description.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -130,9 +130,57 @@ class OnsetPicker extends React.Component {
...
       age: ageValues.age,
       ageUnit: ageValues.ageUnit,
+      year: undefined,
+      month: undefined,
     };
   }
 
+  componentDidMount() {
+    if (this.props.onsetDate) {
+      if (this.props.granularity === 'year' && this.props.onsetDate === moment().year(moment(this.props.onsetDate).year()).startOf('year').format(DATE_FORMAT)) {
```
nramamurth: Please add appropriate code comments or introduce a meaningful variable for `moment().year(moment(this.props.onsetDate).year()).startOf('year').format(DATE_FORMAT))`for better readability.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -141,17 +189,45 @@ class OnsetPicker extends React.Component {
   changeGranularity(granularity) {
     if (granularity === GranularityOptions.AGE) { // Calculate age values and update onsetDate to match age calculation
       this.setState((prevState) => {
-        const ageValues = OnsetUtils.onsetToAge(this.props.birthdate, prevState.onsetDate);
+        if (!!prevState.onsetDate || prevState.granularity !== granularity) {
```
nramamurth: This was probably me while I was prototyping 🤔but this could perhaps now be moved to a separate method that can be invoked whenever values need to be reset.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -182,8 +258,15 @@ class OnsetPicker extends React.Component {
       age = Number(event.target.value);
       this.setState((prevState) => {
       // Check if date can be calculated
-        const ageDate = Number.isInteger(age) && prevState.ageUnit
-          ? moment(this.props.birthdate).add(age, prevState.ageUnit) : undefined;
+        const momentBirthDate = moment(this.props.birthdate);
+        let ageDate = moment(momentBirthDate).add(age, prevState.ageUnit);
+        if (Number.isInteger(age) && prevState.ageUnit) {
+          if ((momentBirthDate.isLeapYear() && !ageDate.isLeapYear() && prevState.ageUnit === AgeUnits.YEARS) || (ageDate.daysInMonth() < momentBirthDate.daysInMonth())) {
```
nramamurth: Documentation 📝

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -268,10 +362,20 @@ class OnsetPicker extends React.Component {
     if (this.props.onsetOnChange === undefined) {
       return;
     }
-
+    const birthDate = moment(this.props.birthdate);
+    if (this.state.onsetDate && moment(this.state.onsetDate) < birthDate) {
+      this.setState(() => ({
+        onsetDate: birthDate.month() !== 11 ? moment().year(birthDate.year()).month(birthDate.month() + 1).startOf('month') : moment().year(birthDate.year() + 1).startOf('year'),
```
nramamurth: 📝

packages/terra-clinical-onset-picker/src/OnsetUtils.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
nramamurth: Why disable?

packages/terra-clinical-onset-picker/src/OnsetUtils.js
```diff
@@ -130,6 +120,38 @@ class OnsetUtils {
     return { age: ageDiff, ageUnit: 'weeks' };
   }
 
+  /**
+   * Converts onset date to a age value with lowest possible age unit (weeks, then months, then years).
```
nramamurth: I'd think that we could stick to having one utility method instead of two - description of `onsetToAge` is the same as that of `initialOnsetToAge` and does not tell when either of them should be leveraged.

nramamurth:
@benbcai @ryanthemanuel Is this a MVB given that the functionality and the requirements are changed?

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Resolved in 1417f94

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Added variables in 1417f94

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Changed in 1417f94

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Added in 1417f94

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Added in 1417f94

packages/terra-clinical-onset-picker/CHANGELOG.md
saket2403: Added in 1417f94

packages/terra-clinical-onset-picker/src/OnsetUtils.js
saket2403: Ooops, removed here 1417f94

packages/terra-clinical-onset-picker/src/OnsetUtils.js
saket2403: Removed the previous one 1417f94

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -108,14 +108,14 @@ class OnsetPicker extends React.Component {
...
+    let ageValues = {};
     if (this.props.ageUnit) {
       ageValues = {
         age: this.props.onsetDate ? moment(this.props.onsetDate).diff(moment(this.props.birthdate), this.props.ageUnit) : undefined,
         ageUnit: this.props.ageUnit,
       };
-    } else {
-      ageValues = OnsetUtils.onsetToAge(this.props.birthdate, moment(this.props.onsetDate));
+    } else if (this.props.onsetDate) {
+      ageValues = OnsetUtils.OnsetToAge(this.props.birthdate, moment(this.props.onsetDate));
```
lokesh-0813: onSetToAge

saket2403: Done 3fb31c1

benbcai:
@nramamurth I don't think we should do a MVB for this change. It is true that there are functional changes but the APIs remain unchanged. Consumers may see some WDIO tests to fail and screenshots would need to be regenerated. In past [PRs](https://github.com/cerner/terra-framework/pull/837#discussion_r330623677) that only had functional/visual impacts but no API changes, we didn't make those as MVB so this should be treated the same. But I interested in hearing what others think.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -268,10 +336,23 @@ class OnsetPicker extends React.Component {
     if (this.props.onsetOnChange === undefined) {
       return;
     }
-
+    const birthDate = moment(this.props.birthdate);
+    // Set onset date to 1st day of next month of birthdate if onset date is less than birthdate.
+    if (this.state.onsetDate && moment(this.state.onsetDate) < birthDate) {
+      this.setState(() => ({
+        onsetDate: birthDate.month() !== 11 ? moment().year(birthDate.year()).month(birthDate.month() + 1).startOf('month') : moment().year(birthDate.year() + 1).startOf('year'),
```
yuderekyu: we might invoke `birthDate.month()`twice. Can we invoke once and save that value into a local variable for re-use before the setState call? We could also apply the same change to `birthDate.year()` for consistency and make this line a tad easier to read :D
```js
const yearOfBirthData = birthDate.year();
const monthOfBirthDate = birthDate.month();
this.setState(() => ({
  onsetDate: monthOfBirthDate !== 11 ? moment().year(yearOfBirthDate).month(monthOfBirthDate + 1).startOf('month') : moment().year(yearOfBirthDate + 1).startOf('year'),
```

saket2403: Added variables ee819ec

ryanthemanuel:
@benbcai I agree.  If the API is the same and we've just fixed the behavior then I'm cool with it not being an MVB.

PaulSolDev:
image

Based on the requirements below:
image

I would expect the day to return: 2001-04-06

PaulSolDev:
Full month range returned in the "less month" scenario.
I would expect this to be locked down to only show available months.

image

saket2403:
Added Filters for months and years in 5474e7e

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -130,28 +130,77 @@ class OnsetPicker extends React.Component {
...
+      month: undefined,
     };
   }
 
+  componentDidMount() {
+    if (this.props.onsetDate) {
+      const firstDayOfYear = moment().year(moment(this.props.onsetDate).year()).startOf('year').format(DATE_FORMAT);
+      const firstDayOfMonth = moment().year(moment(this.props.onsetDate).year()).month(moment(this.props.onsetDate).month()).startOf('month')
+        .format(DATE_FORMAT);
+      if (this.props.granularity === 'year' && this.props.onsetDate === firstDayOfYear) {
```
supreethmr: for consistency `'year'` should be replaced by  `GranularityOptions.YEAR`.

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -130,28 +130,77 @@ class OnsetPicker extends React.Component {
...
       ageUnit: ageValues.ageUnit,
+      year: undefined,
+      month: undefined,
     };
   }
 
+  componentDidMount() {
+    if (this.props.onsetDate) {
+      const firstDayOfYear = moment().year(moment(this.props.onsetDate).year()).startOf('year').format(DATE_FORMAT);
+      const firstDayOfMonth = moment().year(moment(this.props.onsetDate).year()).month(moment(this.props.onsetDate).month()).startOf('month')
```
benbcai: Can we use a variable to store `moment(this.props.onsetDate)` so that it doesn't have to create a new instance every time? And possibly use one for `moment().year(moment(this.props.onsetDate).year())` as well.

benbcai:
@PaulSolDev In the example below where the birthdate is 09/20/2011 with:
granularity: Month
selected year: 2011 (the birth year)

Should the month of September be included as a selectable month since that is the birth month? Or are we excluding September to avoid running into setting the onset date as 2011-09-01, which is before the birthdate?

Screen Shot 2020-02-06 at 12 15 20 PM

PaulSolDev:
@benbcai We would still want September to be a selectable month in that scenario. Good catch.

benbcai:
@PaulSolDev If that's the case, then what should the `day` for onsetDate be when September is chosen as the month. We probably don't want it to be the first day of the month (i.e. 2011-09-01) since that is before the birthdate. Should it be what ever the birthdate is. Likewise, if the granularity is `Year` and 2011 now becomes a year that can be selected, would the onsetDate be `2011-09-20` is 2011 was selected instead of `2011-01-01`?

PaulSolDev:
@benbcai Yes, I think defaulting to the earliest logic date (birthdate) makes sense.

benbcai:
Talked to @PaulSolDev offline regarding whether or not the birth month/year should be excluded from the selection. We decided to excluded it as how it currently works in this PR. If the user wants to select the birth year, the granularity should be fine tuned to `Month`. Likewise, if the user wants to select the birth month, the granularity should be fine tuned to `Date`.

To clarify - no additional code change is needed to address my earlier comment.

packages/terra-clinical-onset-picker/src/OnsetUtils.js
```diff
@@ -9,16 +9,16 @@ const PrecisionOptions = {
...
   /*
    * Create object to pass to SelectField for month options
    * Filters out months before birth or after the current date.
    */
-  static allowedMonths(intl, birthdate, onsetDate) {
+  static allowedMonths(intl, birthdate, year) {
     const start = moment(birthdate);
-    const end = moment();
-    const onsetYear = onsetDate ? onsetDate.year() : moment().year();
+    const end = moment(moment.now());
```
benbcai: What's the reason for passing in `moment.now()`? Is this different from just doing `moment()`?

saket2403: Changed c4b4f2f

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Added Variables c4b4f2f

packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
saket2403: Changed c4b4f2f

PaulSolDev:

Issue #617, opened 2020-03-11
[terra-clinical-result] Numeric value alignment


shinepd:
# Bug Report

## Description
Numeric values are misaligned in the Flowsheet Result Cell

## Steps to Reproduce
1. Create two Flowsheet Result Cells with numeric values with/without normalcy icons present.

## Additional Context / Screenshots

![numeric_alignment](https://user-images.githubusercontent.com/61854307/76432991-b1044700-6381-11ea-8eb6-34080b4f17ad.png)
![margin_highlight](https://user-images.githubusercontent.com/61854307/76432995-b2357400-6381-11ea-8fd7-4a32cd15cf60.png)

## Expected Behavior
All numeric values should be aligned on the left side of the cell, accounting for the normalcy icon.

## Environment

* Component Name and Version: terra-clinical-result@1.0.0
* Browser Name and Version: Chrome 80 / IE 11
* Node/npm Version: Node 8 / npm 6.13.4
* Webpack Version: ^4.41.5
* Operating System and version (desktop or mobile): desktop

shinepd:
@mjhenkes as a consumer, we'd also like the ability to display a checkbox in the cell for selection of unverified values. This selection is independent of the pre-existing cell selection. As unverified values will not have a normalcy icon present, ideally there would be vertical alignment between the normalcy icon and checkbox. However, in order to accomplish this we'd need to know the width of the normalcy icon prior to rendering. If it makes more sense to set something up to chat about this, please let me know and I can schedule something.

mjhenkes:
@shinepd Can you log a new issue for your request? (I'm assuming its a new request) We can discuss it there. This was closed because it was fixed (afaik) with PR #620.

Issue #620, opened 2020-03-11
[terra-clinical-result] Multiple Value Indicator color not displaying


shinepd:
# Bug Report

## Description
After the [interpretationPropTypes](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-result/src/proptypes/interpretationPropTypes.js) were updated to be lower cased, the multiple value indicator is no longer showing color when a face-down value is critical.

## Steps to Reproduce
1. Create a Flowsheet Result Cell with multiple values.
2. Ensure at least one of the face-down values is critical.

## Additional Context / Screenshots

![numeric_alignment](https://user-images.githubusercontent.com/61854307/76434183-579d1780-6383-11ea-94e4-709e30d33f38.png)

## Expected Behavior
For values that are critical, the color of the multiple value indicator should change.

## Possible Solution
If the interpretation prop is to be treated as an enumerable value, I think values should be all upper cased or at least not be strict on lower casing.

## Environment

* Component Name and Version: terra-clinical-result@1.0.0
* Browser Name and Version: Chrome 80 / IE 11
* Node/npm Version: Node 8 / npm 6.13.4
* Webpack Version: ^4.41.5
* Operating System and version (desktop or mobile): desktop

BenBoersma:
This seems to be caused by the interpretation prop being given an invalid value. As such closing the issue as not an issue.

shinepd:
It's actually due to terra strictly enforcing the wrong casing for enumerable values, forcing consumers to write a translator function.

mjhenkes:
@shinepd, What reference are you using to cite the correct casing of javascript enums? All terra enums are lowercase throughout our api, we are remaining consistent with ourselves.

StephenEsser:
@shinepd 

What is the right casing for these values? Terra component Proptypes have always been case sensitive and are usually always lowercase.

PolymorphedSquirrel:
Take a peek at this where the normalcy is provided after going to lowercase.  An interpretation of 'normal' is supplied to the FlowsheetResultCell for a normal result.

![image](https://user-images.githubusercontent.com/43280824/76792523-3aa48200-6791-11ea-8e1d-9edf88d97315.png)

When the result is not unverified and is supplied with the lowercase value, the resulting display is not retaining the space an icon would take up if present.

[...]

PR #574, opened 2019-09-26
Clinical Result 

neilpfeiffer:
### Summary

New components needed to display results in flowsheets, detail views, etc.
List includes:
- _Observation base template
- Clinical Result
- Blood Pressure Result
- Flowsheet Cell (multiple results)
- Result Time Header Cell
- Result Name Header Cell

In Development - including the new components as they are stable and ready for review, still need to add tests.

Deployment:
https://terra-clinical-deployed-pr-574.herokuapp.com/components/terra-clinical-result/clinical-result/clinical-result

Resolves: #576

PR #621, opened 2020-03-12
[terra-clinical-result] Fixing issues

BenBoersma:
### Summary
*Edit Removed the reference to issue 620
Fixes https://github.com/cerner/terra-clinical/issues/617
Adding defensive coding to prevent the above bugs from happening.

@cerner/terra


neilpfeiffer:
Reviewed: 
UX +1

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
```diff
@@ -281,11 +291,11 @@ const createFlowsheetResultCellDisplay = (resultDataSet, hideUnit, numericOverfl
     additionalResultList.forEach((result) => {
       const { isSingleResult, isPairedResult } = checkIfSingleOrPairedResult(result);
       if (isSingleResult) {
-        const resultInterpretation = !isEmpty(result.interpretation) && !result.isUnverified ? result.interpretation : null;
+        const resultInterpretation = !isEmpty(result.interpretation) && !result.isUnverified ? result.interpretation.trim().toLowerCase() : null;
```
yuderekyu: https://github.com/cerner/terra-clinical/pull/574#discussion_r386189233

petition to replace this isEmpty usage anywhere in the component with the double bang

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
```diff
@@ -215,7 +225,7 @@ const unpackResultAttributes = (resultDataItem) => {
     isUnverified,
   } = resultDataItem;
   const itemAttributes = new AttributesTemplate();
-  itemAttributes.interpretation = interpretation;
+  itemAttributes.interpretation = interpretationsWithIcons.includes(interpretation);
```
emilyrohrbough: Is `itemAttributes.interpretation` suppose to be a true or false value?

packages/terra-clinical-result/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ ChangeLog
 
 Unreleased
 -----------------
+### Fixed
+* Fixed an alignment issue with FlowsheetResultCell that caused normal interpretation results to be aligned to the left.
+
+### Added
+* Added additional defensive coding to ensure the multiple result icon is colored correctly according to interpretation.
```
emilyrohrbough: What change is this one?

packages/terra-clinical-result/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ ChangeLog
 
 Unreleased
 -----------------
+### Fixed
+* Fixed an alignment issue with FlowsheetResultCell that caused normal interpretation results to be aligned to the left.
```
emilyrohrbough: which change affected the alignment issue? Is this from trimming the interpretation value?

emilyrohrbough:
Was there missing screenshots previously or was there a test change I missed?

Should there be some sort of jest or wdio test added to validate these changes?

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
neilpfeiffer: Can't -- Linter doesn't like it how we have it currently written as the comparison side of the ternaries.  https://eslint.org/docs/rules/no-extra-boolean-cast. (It's fine to double cast when setting/returning values to new variables, but not fine to double-cast for bool comparisons.)

+1 to the Suggestion and https://github.com/cerner/terra-clinical/pull/574#discussion_r386189233 - better option we've talked about is probably refactor to remove these entirely -- one idea is an immediate result object and  initializer function to detect all of  the prop object and string checks like these up front, and then we can just utilize straight booleans downstream throughout the rest of the project.

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
BenBoersma: It ends up becoming used as boolean later in `createStandardResultDisplay` and `createBloodPressureResultDisplay`. Other then that, I didn't find another place where it being a boolean adversely affected what we were using it for.

packages/terra-clinical-result/CHANGELOG.md
BenBoersma: These https://github.com/cerner/terra-clinical/pull/621/files#diff-a81e0ffc96673dad2a43e3b9b81f7090R294 
A user was still inputting a non-lowercase value for interpretation in, which was causing us to not display the correct interpretation color for the extra results icon.

packages/terra-clinical-result/CHANGELOG.md
BenBoersma: For the alignment to be displayed correctly, we add a class when there is no icon which places a margin to line up the text as if there was an icon. We were mistaking the "normal" interpretation to be one that provided an icon and thus weren't providing that class. The inclusion of [this array](https://github.com/cerner/terra-clinical/pull/621/files#diff-a81e0ffc96673dad2a43e3b9b81f7090R63) and the changes to see if an interpretation are in that array are how this PR fixes that issue.

BenBoersma:
> Was there missing screenshots previously or was there a test change I missed?
> 
> Should there be some sort of jest or wdio test added to validate these changes?

https://github.com/cerner/terra-clinical/blob/3cae2b5a093df45a6eca677e04ba074e1b1c939a/packages/terra-clinical-result/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_medium/flowsheet-result-cell-spec/interpretations%5Bdefault%5D.png
This test shows the alignment issues, it was just passed over during the first release.

As for the multiple result icon, it was more of an issue with a consumer using the prelowercase API and the defensive coding is more to ensure standardization since we do similar trim to lowercase data manipulation elsewhere.

packages/terra-clinical-result/CHANGELOG.md
emilyrohrbough: @BenBoersma our API specifies lower-case: https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-result/src/proptypes/observationPropTypes.js#L17

 I would say that's user error. we don't do update props strings in other components to be upper or lower case

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
neilpfeiffer: yes - this item attributes object was the start of doing an initialization to save off a series of bools to then use later for quicker yes/no checks.

Could rename object and keys to something like 'hasAttributes.interpretationWithIcon" to be better? Something like:

```suggestion  
const hasAttributes = new AttributesTemplate();
hasAttributes.interpretationWithIcon = interpretationsWithIcons.includes(interpretation);
hasAttributes.comment = hasComment;
hasAttributes.modified = isModified;
hasAttributes.unverified = isUnverified;
```

emilyrohrbough: Is it possible to update this key to `hasInterpretationIcon` to indicate it's a boolean value? Is there any downstream impact making this change or is is only used in this component?

packages/terra-clinical-result/CHANGELOG.md
emilyrohrbough: Thanks for the info!

neilpfeiffer: No. This adds extra space on the left to any value which doesn't have an interpretation that also displays an icon, so that all values, icon or without icon, all vertically align when in a large flowsheet grid.

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
neilpfeiffer: rather than the just the one key to have `has`, since all the key:values are boolean, rename to something like this?

``` 
const hasAttributes = new AttributesTemplate();
hasAttributes.interpretationIcon = interpretationsWithIcons.includes(interpretation);
hasAttributes.comment = hasComment;
hasAttributes.modified = isModified;
hasAttributes.unverified = isUnverified;
```

packages/terra-clinical-result/CHANGELOG.md
neilpfeiffer: +1. Since pre-release versions of the component had this prop enums as all uppercase, is a bit of generosity for current consumers until they can also update to match the final/updated API.

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
yuderekyu: i'm confused, would a simple existence check suffice then?
```js
const resultInterpretation = result.interpretation && !result.isUnverified ? result.interpretation.trim().toLowerCase() : null;
```

packages/terra-clinical-result/CHANGELOG.md
mjhenkes: Lets not do this. Like @emilyrohrbough said we don't do this for any other string value.

BenBoersma: As per scrum this morning, I have removed the `trim(). toLowerCase()` calls. https://github.com/cerner/terra-clinical/pull/621/commits/7ddb8bdb5bd8b30a3be7567409ab24d756f6ec83#diff-a81e0ffc96673dad2a43e3b9b81f7090L294

emilyrohrbough: @BenBoersma Does this changelog entry need removed then?

BenBoersma: Removed now https://github.com/cerner/terra-clinical/pull/621/commits/1222b6919583cd9c37c03c7b3c99a8d1b032cca7

emilyrohrbough:
@BenBoersma is this issue resolved still or does it need to be closed as a non-issue?
https://github.com/cerner/terra-clinical/issues/620

BenBoersma:
> @BenBoersma is this issue resolved still or does it need to be closed as a non-issue?
> #620

Probably closed as a non-issue since we've clarified as user error.

emilyrohrbough:
@BenBoersma Sounds good on updating the issue and removing the reference that this PR fixes it.

PR #641, opened 2020-04-08
[terra-clinical-header] Update Header Styles to Match terra-action-header

### LINKED ISSUE #640: [clinical-header] Update Fusion Theme Styles
# Feature Request

## Description
Update Fusion Theme Styles for terra-clinical-header. These styles should be consistent with the latest terra-action-header changes in https://github.com/cerner/terra-core/pull/2714.

## @ Mentions
I will be working with @jeremyfuksa for these values.

emilyrohrbough:
### Summary
Update Header Styles to Match terra-action-header. This has some small tweaks to the default-theme and the fusion theme.

I updated the background color of the fake content divs so it's visible in low-light theme since black on black is hard to see 😄 

This is using the new terra-theme-context! yay! 🎉 

Closes #640 

### Deployment Link

https://terra-clinic-header-sty-6j2809.herokuapp.com/home

### Testing

- exiting jest & wdio tests

### Additional Details

Directions were to match terra-action-header styles made on https://github.com/cerner/terra-core/pull/2714.

@cerner/terra


packages/terra-clinical-header/package.json
```diff
@@ -5,7 +5,8 @@
   "description": "A Header component that allows elements to be placed on the left and right ends of the header with a left aligned title in the center. ",
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/cerner/terra-clinical.git"
+    "url": "git+https://github.com/cerner/terra-clinical.git",
+    "dir": "packages/terra-clinical-header"
```
emilyrohrbough, line 9: for coming changes: https://github.com/cerner/terra-dev-site/pull/243

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -44,10 +47,7 @@ const defaultProps = {
 const Header = ({
   children, title, startContent, endContent, isSubheader, ...customProps
 }) => {
-  let startElement;
-  if (startContent) {
-    startElement = <div className={cx('flex-end')}>{startContent}</div>;
-  }
+  const theme = useContext(ThemeContext) || {};
```
emilyrohrbough, line 50: I'm falling back to `{}` for any components rendered in a project that does not have the context defined

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -60,30 +60,29 @@ const Header = ({
...
-    childElement = React.cloneElement(children, { className: childClassNames });
-  }
+  const content = React.Children.map(children, child => (
+    React.cloneElement(child, {
+      className: classNames([cx('flex-collapse'), children.props.className]),
+    })
+  ));
 
-  const headerClass = isSubheader ? 'flex-subheader' : 'flex-header';
+  const headerClassNames = classNames([
```
emilyrohrbough, line 69: the classNames module will filter out any class that is undefined, which is why I am adding `customProps.className` directly. I used `classNames` instead of `cx` to prevent binding the custom class to our stylesheet.

packages/terra-clinical-header/src/Header.module.scss
```diff
@@ -1,77 +1,80 @@
...
+  .flex-end + .flex-fill {
+    .title {
+      padding-left: var(--terra-clinical-header-end-content-plus-title-padding-left, 0.35714rem);
+    }
+  }
+  
+  .flex-fill + .flex-end {
+    .title {
+      padding-right: var(--terra-clinical-header-title-plus-end-content-padding-right, 0.35714rem);
+    }
```
emilyrohrbough, line 77: @jeremyfuksa can you config these changes are desired to match: https://github.com/cerner/terra-core/pull/2714/files?file-filters%5B%5D=.md&file-filters%5B%5D=.png&file-filters%5B%5D=.scss&file-filters%5B%5D=.snap#diff-bf4aa35fd6c961609389f4c3de68798bR62-R67

I assume we wanted to pad both sides since we don't have knowledge on what content is being placed in the start/end content boxes. This isn't needed for terra-action-header since it controls the padded styles on the button content it drops in here: https://github.com/cerner/terra-core/blob/master/packages/terra-action-header/src/ActionHeader.module.scss#L4

packages/terra-clinical-header/src/themes/clinical-lowlight-theme/Header.module.scss
```diff
@@ -0,0 +1,10 @@
+:local {
```
emilyrohrbough, line 1: @mjhenkes what are your thoughts on organizing themeing this way so file names are cleaner. If a component has multiple files, we can following the File naming like:
```
Main.jsx
Main.module.scss
Another.jsx
Another.module.scss
/themes
   /some-theme
       Main.moudle.scss
       Another.module.scss
```

packages/terra-clinical-header/src/themes/orion-fusion-theme/Header.module.scss
```diff
@@ -0,0 +1,34 @@
+:local {
+  .orion-fusion-theme {
+    // .flex-header
+    --terra-clinical-header-align-items: center;
+    --terra-clinical-header-background-color: #f4f4f4;
+    --terra-clinical-header-border-bottom: 0.16666667rem solid #64696c;
```
emilyrohrbough, line 6: @jeremyfuksa you mentioned this style was correct, which is actually incorrect in terra-action-header: https://github.com/cerner/terra-core/pull/2714/files?file-filters%5B%5D=.md&file-filters%5B%5D=.png&file-filters%5B%5D=.scss&file-filters%5B%5D=.snap#diff-88dce3fa259e93fb4378f904ec7774c0L5

packages/terra-clinical-header/src/themes/orion-fusion-theme/Header.module.scss
```diff
@@ -0,0 +1,34 @@
...
+    --terra-clinical-header-min-height: 2.83333rem;
+    --terra-clinical-header-padding: 0;
+
+    // .flex-collapse
+    --terra-clinical-header-flex-collapse-align-self: center;
+    --terra-clinical-header-flex-collapse-margin: 0.5rem;
+
+    // .subheader
+    --terra-clinical-header-subheader-background-color: #f4f4f4;
+    --terra-clinical-header-subheader-min-height: 2.83333rem;
```
emilyrohrbough, line 17: @jeremyfuksa I guess on this value. By not providing a value, the default min-height is larger than the regular min-height value for this theme. Can you validate or provide an updated value for this?

packages/terra-clinical-header/src/Header.jsx
mjhenkes: Can the context do that by default? I was thinking thats what this would do:
https://github.com/cerner/terra-framework/blob/master/packages/terra-theme-context/src/ThemeContext.jsx#L4

packages/terra-clinical-header/src/Header.jsx
```diff
@@ -1,9 +1,12 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames/bind';
+import classNames from 'classnames';
```
mjhenkes, line 3: I like what you're doing with classnames here

packages/terra-clinical-header/src/Header.jsx
emilyrohrbough: OH I didn't realize there was a default. I'll assume it works fine then. 👍

packages/terra-clinical-header/src/themes/clinical-lowlight-theme/Header.module.scss
mjhenkes: I could see it, adding the name to every file is a bit annoying. What if we dropped the theme folder?:
```
Main.jsx
Main.module.scss
Another.jsx
Another.module.scss
 /some-theme
     Main.moudle.scss
     Another.module.scss
```

emilyrohrbough: I'm fine with dropping the theme folder. So far each theme is name `something-theme`. If we follow this pattern, it describes itself

mjhenkes: 👍 I'll update my terra-core pr updating button and card to the same pattern.

packages/terra-clinical-header/CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* Added orion-fusion-theme style values
+
+### Changes
```
StephenEsser, line 9:
```suggestion
### Changed
```

https://engineering.cerner.com/terra-ui/contributing/terra-ui/developer-workflow#updating-the-changelog

packages/terra-clinical-header/src/Header.module.scss
```diff
@@ -1,77 +1,79 @@
...
-    overflow: hidden;
-    padding: var(--terra-clinical-header-padding, 0.714rem);
+    min-height: var(--terra-clinical-header-min-height, 2.92857rem);
+    padding: var(--terra-clinical-header-padding, 0.35714rem 0.35714rem 0.35714rem 0);
     position: relative;
     width: 100%;
 
     .flex-collapse {
-      align-items: center;
+      align-self: var(--terra-clinical-header-flex-collapse-align-self);
```
StephenEsser, line 19: Does this need a default fallback or is it only used within other themes?

emilyrohrbough: This is only used by the fusion-theme

packages/terra-clinical-header/src/Header.module.scss
```diff
@@ -1,77 +1,80 @@
...
     align-self: center;
     flex: 1 1 auto;
-    height: 100%;
-    min-width: 0;
     position: relative;
     width: 100%;
   }
 
   .flex-end {
-    align-items: center;
```
jeremyfuksa, line 35: This needs to be here, otherwise, the buttons are butted up against the top of the box.

packages/terra-clinical-header/src/Header.module.scss
jeremyfuksa: I think you're right in padding both sides.

packages/terra-clinical-header/src/themes/orion-fusion-theme/Header.module.scss
jeremyfuksa: Yeah. Oops. Let's stick with `9b9fa1` as it is in action-header.

jeremyfuksa: Oops. I told you the wrong one. `2px solid #9b9fa1` was it.

packages/terra-clinical-header/src/Header.module.scss
emilyrohrbough: @jeremyfuksa 

should https://github.com/cerner/terra-clinical/pull/641/files#diff-9f987196835c1f499bb09273c8f5325dR7 be updated from `flex-start` to `center` then?

is action-header incorrect? 
- [action-header header align-items style](https://github.com/cerner/terra-core/blob/master/packages/terra-action-header/src/ActionHeaderContainer.module.scss#L3)
- [action-header flex-end align-self style](https://github.com/cerner/terra-core/blob/master/packages/terra-action-header/src/ActionHeaderContainer.module.scss#L36)

packages/terra-clinical-header/src/themes/orion-fusion-theme/Header.module.scss
emilyrohrbough: When comparing the fusion component & nav header, their value in the dom is `0.16666667rem solid #64696c` for the bottom border.

packages/terra-clinical-header/src/themes/orion-fusion-theme/Header.module.scss
jeremyfuksa: Since there's no actual Fusion equivalent to a subheader for this, I think keeping the two types at the same height is OK for now. If we get a new design direction from Seth down the line, we can address then.

packages/terra-clinical-header/src/themes/orion-fusion-theme/Header.module.scss
jeremyfuksa: Oh, yes. I see what you're saying. The border should match Fusion, so it looks like we need to update the theme in action header as well.

packages/terra-clinical-header/src/themes/orion-fusion-theme/Header.module.scss
emilyrohrbough: sounds good. I will update the value to match

packages/terra-clinical-header/src/Header.module.scss
jeremyfuksa: They aren't incorrect per se, they're just coded in a way that the inner content is compensating (via line-height, etc.) for what the button doesn't. The text in the header is fine with its alignment style. It's just the buttons that need flexbox vertical alignment.

emilyrohrbough: updated: https://github.com/cerner/terra-clinical/pull/641/commits/daeb48597a485300015085b1bccc2a5725d42829

packages/terra-clinical-header/src/Header.module.scss
emilyrohrbough: update backed on feedback from Jeremy: https://github.com/cerner/terra-clinical/pull/641/commits/daeb48597a485300015085b1bccc2a5725d42829

packages/terra-clinical-header/src/Header.module.scss
jeremyfuksa: Screen Shot 2020-04-09 09 31 35 AM

Still seeing the buttons butting up against the top.

emilyrohrbough: fixed: https://github.com/cerner/terra-clinical/pull/641/commits/0269a181756802e99eac7be85fbbe98734dd096b

PR #642, opened 2020-04-08
Condense wdio config

emilyrohrbough:
### Summary

There are a lot of wdio configuration in this project. It was a little confusing to me to keep track of which config was being used or not. I took the existing package-level wdio logic and pushed to a root wdio level.

If people are against these changes, that's okay, but seem like this could be simplified a bit.

I also add npm run test to the package-level test scripts.

### Deployment Link

https://terra-clinic-condense-w-hk6yac.herokuapp.com/

### Testing
running tests locally + travis builds

### Additional Details

@cerner/terra


emilyrohrbough:
@yuderekyu so you can run wdio from the package-level for those tests so you didn't have to run all wdio tests in the repo

PR #648, opened 2020-04-16
[terra-clinical-result] Cleaning up clinical-result

BenBoersma:
### Summary

Cleaning up clinical-result, mostly addressing leftover comments from this PR: https://github.com/cerner/terra-clinical/pull/574
Specifically
* Changed the description of 'resultNoData' prop for observation prop types
* Changed ClinicalResultBloodPresure to return early if result has an Error or doesn't have data
* Added to interpretation Bloodpressure test
* Added single result bloodpressure test
* Added Result hasError test
* Added Result hasNoData test
* Added Decorated Flowsheet test
* Added Flowsheet hasError test
* Added Flowsheet hasNoData test
* Added Flowsheet numericOverflow test
* Removed unused test data
* Removed default empty object prop from ClinicalResult and Observation.

### Deployment Link

https://terra-clinical-deployed-pr-#.herokuapp.com/

### Additional Details

@cerner/terra


packages/terra-clinical-result/src/common/observation/_Observation.jsx
```diff
@@ -69,7 +65,7 @@ const Observation = (props) => {
     ...customProps
   } = props;
 
-  const isValidValue = !!result.value;
+  const isValidValue = result && !!result.value;
```
emilyrohrbough, line 68: There are a few more places in this file were we will want to valid result has been provided.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -120,144 +120,144 @@ const ClinicalResultBloodPressure = (props) => {
...
-  } else {
-    const compareUnits = {};
-    const compareStatusIsInError = {};
+    return <ResultError />;
+  }
+  if (hasResultNoData) {
+    return <NoData />;
+  } 
+  const compareUnits = {};
+  const compareStatusIsInError = {};
```
emilyrohrbough, line 130: I think it'd be work creating two new variables for systolic & diastolic with fallback to {} so we can remove all the extra `systolic && systolic.{key}` checks in this file to ensure systolic is not undefined 
```
const systolicResult = systolic || {};
const diastolicResult = diastolic || {};
```

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -120,144 +120,144 @@ const ClinicalResultBloodPressure = (props) => {
...
-      if (!isEmpty(result.unit)) { compareUnits.diastolic = result.unit.trim().toLowerCase(); }
-      if (!isEmpty(status)) { compareStatusIsInError.diastolic = checkIsStatusInError(status); }
-      if (!isEmpty(conceptDisplay)) { compareConceptDisplays.diastolic = conceptDisplay.trim().toLowerCase(); }
-      if (!isEmpty(datetimeDisplay)) { compareDatetimeDisplays.diastolic = datetimeDisplay.trim().toLowerCase(); }
-    }
+    if (!isEmpty(result.unit)) { compareUnits.diastolic = result.unit.trim().toLowerCase(); }
+    if (!isEmpty(status)) { compareStatusIsInError.diastolic = checkIsStatusInError(status); }
+    if (!isEmpty(conceptDisplay)) { compareConceptDisplays.diastolic = conceptDisplay.trim().toLowerCase(); }
+    if (!isEmpty(datetimeDisplay)) { compareDatetimeDisplays.diastolic = datetimeDisplay.trim().toLowerCase(); }
+  }
```
emilyrohrbough, line 170: this is 1:1 with the `if (systolic && !noDataSystolic)`. Can we bump this into some some sort of utils file so we can write jest tests on this logic & make it reusable?

packages/terra-clinical-result/tests/jest/__snapshots__/ClinicalResultBloodPressure.test.jsx.snap
```diff
@@ -862,21 +862,9 @@ exports[`ClinicalResultBloodPressure - Systolic - should render with no data 1`]
...
->
-  <InjectIntl(KnownNoData) />
-</div>
-`;
+exports[`ClinicalResultBloodPressure should render a NoData if hasResultNoData is true 1`] = `<InjectIntl(KnownNoData) />`;
 
-exports[`ClinicalResultBloodPressure should render a ResultError if hasResultError is true 1`] = `
-<div
-  className="clinical-result blood-pressure-result"
->
```
emilyrohrbough, line 876: Do we want to drop this wrapper?

packages/terra-clinical-result/src/common/observation/_Observation.jsx
BenBoersma: I don't think in this file? The rest of the calls for `result` in this file are in an `if` that uses `isValidValue`.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -120,144 +120,144 @@ const ClinicalResultBloodPressure = (props) => {
...
-        );
-        decoratedResultDisplay.push(diastolicDisplay);
-      }
+  const decoratedResultDisplay = [];
+  if (systolic || diastolic) {
+    if (!systolic) {
+      decoratedResultDisplay.push(<ResultError key={`Error-Systolic-${id}`} />);
+    } else if (noDataSystolic) {
+      decoratedResultDisplay.push(<NoData key={`NoData-Systolic-${id}`} />);
+    } else if (systolic) {
```
yuderekyu, line 184: this block is massive :( would it make more sense to create helpers based on systolic and diastolic? we can make the code easier to read. something similar to below:

```js
/* helpers. first two could be further reduced by another helper */
const createSystolicDisplay(systolic) {
  if (systolic) { 
    // return wrapper
  } else if (systolic.noDataSystolic) { 
    // return no data
  } else { 
    // return error 
  }
}

const createSystolicDisplay(diastolic) {
  if (diastolic) { 
    // return wrapper
  } else if (diastolic.noDataDiastolic) { 
    // return no data
  } else { 
    // return error 
  }

const createResultDisplay = (systolic, diastolic) => {
  if (!systolic && !diastolic) {
    return;
  } else {
    const resultDisplay = [];
    resultDisplay.push(createSystolicDisplay(systolic);
    resultDisplay.push(separator);
    resultDisplay.push(createDiastolicDisplay(systolic);
    return resultDisplay;
  }

...
/* use helper */
const resultDisplay = createResultDisplay(systolic, diastolic);
```

what are your thoughts?

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Added in the object, https://github.com/cerner/terra-clinical/pull/648/commits/a037421a0680c7296aa23625adb73abd2787d211
this might change as it looks like i'm gonna retool this section in general.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: First draft at changing the functions, https://github.com/cerner/terra-clinical/pull/648/commits/5611e410a5b4c6389abf03f7d50fbb975c73143e#diff-72ca9008b342e1caa5acc1466dfa7f20R33
Still need to add some tests but didn't want to go down that rabbit hole until i got feedback.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Split it up, but didn't make the createResultDisplay function, because we have a few others things that get set under that check.
https://github.com/cerner/terra-clinical/pull/648/commits/5611e410a5b4c6389abf03f7d50fbb975c73143e#diff-13926e711e02256769a4612c0d911268R106

packages/terra-clinical-result/src/common/utils.js
```diff
@@ -25,6 +25,34 @@ const checkTypeNumeric = (resultData) => {
...
+/*
+ * Returns a copy of the object with additional parameters with trimmed values and informational booleans
+ * @param {object} observationPropShape
+ * @return {object} is an object with added trimmed values and information.
+ */
+const createResult = (data) => {
+  if(!data) {
+    return {};
+  }
+  const returnResult = JSON.parse(JSON.stringify(data));
```
emilyrohrbough, line 37: Could we use data directly without JSON.parse & JSON.stringify?

packages/terra-clinical-result/src/common/utils.js
```diff
@@ -25,6 +25,34 @@ const checkTypeNumeric = (resultData) => {
...
+  returnResult.noData = (data.resultNoData === true);
+  if (!returnResult.noData) {
+    const {
+      result,
+      status,
+      conceptDisplay,
+      datetimeDisplay,
+    } = data;
+
+    if (!isEmpty(result.unit)) { returnResult.cleanedUnit = result.unit.trim().toLowerCase(); }
```
emilyrohrbough, line 48:
```suggestion
    if (!isEmpty(result.unit)) {
      returnResult.cleanedUnit = result.unit.trim().toLowerCase();
    }
```

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -9,7 +9,12 @@ import Observation from './common/observation/_Observation';
 import observationPropShape from './proptypes/observationPropTypes';
 import ResultError from './common/other/_ResultError';
 import NoData from './common/other/_KnownNoData';
-import { isEmpty, checkIsStatusInError, ConditionalWrapper } from './common/utils';
+import {
+  isEmpty,
```
emilyrohrbough, line 13: can we remove isEmpty?

packages/terra-clinical-result/CHANGELOG.md
```diff
@@ -3,6 +3,23 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Changed the description of 'resultNoData' prop for observation prop types
+* Changed ClinicalResultBloodPresure to return early if result has an Error or doesn't have data
```
emilyrohrbough, line 8:
```suggestion
* Changed ClinicalResultBloodPressure to return early if result has an Error or doesn't have data
```

packages/terra-clinical-result/src/common/utils.js
```diff
@@ -25,6 +25,34 @@ const checkTypeNumeric = (resultData) => {
   return isNumeric;
 };
 
+/*
+ * Returns a copy of the object with additional parameters with trimmed values and informational booleans
+ * @param {object} observationPropShape
+ * @return {object} is an object with added trimmed values and information.
+ */
+const createResult = (data) => {
```
emilyrohrbough, line 33: Should we rename to `santatizeResult`? I was a littl econfused when reading through Blood pressure and had to back track a bit to understand the difference between this object and the original

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -103,6 +108,58 @@ const createDatetimeDisplays = (compareDatetimeDisplays) => {
...
+      condition={diastolic.statusInError}
+      wrapper={children => <del>{children}</del>}
+    >
+      <Observation
+        key={`Observation-Diastolic-${diastolic.eventId}`}
+        eventId={diastolic.eventId}
+        result={diastolic.result}
+        interpretation={!diastolic.statusInError ? diastolic.interpretation : null}
+        isUnverified={diastolic.isUnverified}
+        hideUnit={hideUnit}
```
emilyrohrbough, line 156: From what I can tell, it seems the biggest differences between `createSystolicDisplay` and `createDiastolicDisplay` are
- the use of the word `Systolic ` vs `Diastolic` in the key names
- how the hideUnit value is handled. 

1. could we combine these to be resuable and take a `type` parameter?
2. would it make sense to break his out into it's own class so we can validate the rendered component in testing? There are ~5-6 combinations that affect how the Systolic result  renders and 4 combos rendering variants for diastolic result.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -120,144 +177,73 @@ const ClinicalResultBloodPressure = (props) => {
...
+  const hasUnverifiedIcon = (systolicResult.isUnverified) || (diastolicResult.isUnverified);
 
-      if (!isEmpty(result.unit)) { compareUnits.diastolic = result.unit.trim().toLowerCase(); }
-      if (!isEmpty(status)) { compareStatusIsInError.diastolic = checkIsStatusInError(status); }
-      if (!isEmpty(conceptDisplay)) { compareConceptDisplays.diastolic = conceptDisplay.trim().toLowerCase(); }
-      if (!isEmpty(datetimeDisplay)) { compareDatetimeDisplays.diastolic = datetimeDisplay.trim().toLowerCase(); }
-    }
+  let iconGroupDisplayElement = null;
+  const decoratedResultDisplay = [];
+  if (systolic || diastolic) {
```
emilyrohrbough, line 209: should we have checked that  at least systolic or diastolic was provided above and if not, should we have` returned <NoData />;` ?

emilyrohrbough: Also, should we update this to use `systolicResult` and `diastolicResult` since these are the sanitized result objects we are using to render content with?

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -120,144 +177,73 @@ const ClinicalResultBloodPressure = (props) => {
...
+    decoratedResultDisplay.push(createSystolicDisplay(systolicResult, hideUnit, diastolicResult.cleanedUnit, id));
+    decoratedResultDisplay.push(<span key={`Observation-Separator-${(systolic) ? systolic.eventId : diastolic.eventId}`} className={cx('result-display-separator')}>/</span>);
+    decoratedResultDisplay.push(createDiastolicDisplay(diastolicResult, hideUnit, id));
```
emilyrohrbough, line 212:
```suggestion
    decoratedResultDisplay = (
      <>
        {createSystolicDisplay(systolicResult, hideUnit, diastolicResult.cleanedUnit, id)}
        <span key={`Observation-Separator-${(systolic) ? systolic.eventId : diastolic.eventId}`} className={cx('result-display-separator')}>/</span>
        {createDiastolicDisplay(diastolicResult, hideUnit, id)}
      </>
```

packages/terra-clinical-result/src/common/observation/_Observation.jsx
emilyrohrbough: you right 👍

packages/terra-clinical-result/src/common/utils.js
```diff
@@ -35,17 +63,10 @@ const checkTypeNumeric = (resultData) => {
...
-  checkTypeNumeric,
-  ConditionalWrapper,
-};
-
-export default ClinicalResultUtils;
 export {
   isEmpty,
   checkIsStatusInError,
   checkTypeNumeric,
+  createResult,
```
emilyrohrbough, line 70: jest test on this guy?

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -120,144 +177,73 @@ const ClinicalResultBloodPressure = (props) => {
...
+  });
 
-      if (!isEmpty(result.unit)) { compareUnits.systolic = result.unit.trim().toLowerCase(); }
-      if (!isEmpty(status)) { compareStatusIsInError.systolic = checkIsStatusInError(status); }
-      if (!isEmpty(conceptDisplay)) { compareConceptDisplays.systolic = conceptDisplay.trim().toLowerCase(); }
-      if (!isEmpty(datetimeDisplay)) { compareDatetimeDisplays.systolic = datetimeDisplay.trim().toLowerCase(); }
-    }
+  const datetimeDisplayElement = createDatetimeDisplays({
+    originalSystolic: systolicResult.datetimeDisplay,
+    originalDiastolic: diastolicResult.datetimeDisplay,
```
emilyrohrbough, line 198: do we need to check against the original values? I would have expected the cleaned values would be equivalent... It might even be better for this check: https://github.com/cerner/terra-clinical/blob/86ab6dbbc5f074c20783c6f4830136bb7a43f63c/packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx#L93
Incase there's trailing whitespace

emilyrohrbough: actually nvm, in the method I linked, systolic and diastolic do not map to the original objects

packages/terra-clinical-result/src/common/utils.js
BenBoersma: if we use `data` directly we start adding to the original data source of `systolic` and `diastolic`. I was trying to find a way to clone the object that worked in all browsers.

emilyrohrbough: did you try `{ ...data }`?

BenBoersma: Well now I feel dumb......
Yea that works.

packages/terra-clinical-result/CHANGELOG.md
```diff
@@ -3,6 +3,23 @@ ChangeLog
...
+* Added Result hasError test
+* Added Result hasNoData test
+* Added Decorated Flowsheet test
+* Added Flowsheet hasError test
+* Added Flowsheet hasNoData test
+* Added Flowsheet numericOverflow test
+
+### Removed
+* Removed unused test data
+* Removed default empty object prop from ClinicalResult and Observation.
```
StephenEsser, line 23: Woah, this is a pretty hefty list of changes in one pull request

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -103,6 +108,58 @@ const createDatetimeDisplays = (compareDatetimeDisplays) => {
   return null;
 };
 
+const createSystolicDisplay = (systolic, hideUnit, diastolicUnit, id) => {
+  if (!Object.prototype.hasOwnProperty.call(systolic, 'noData')) {
```
StephenEsser, line 112: This is a pretty verbose logic statement. Is this doing more than just checking if the property exists?

```js
if (!systolic.noData)
```

BenBoersma: Originally this `if` was to check if the component's `systolic` prop was defined and if not return `ResultError`. With the new way we define things the passed `systolic` object is actually our custom object that adds in a few more trimmed values and booleans but is always defined to avoid having to prepend `systolic &&` to a lot of if statements.  However this custom object might be an empty object if the `systolic` prop wasn't defined.  It's `noData` property always will be defined( and might be false) if the prop was though.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Tests added https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-8b312db510259892da0886d2f909d49eR89

packages/terra-clinical-result/src/common/utils.js
BenBoersma: Fixed now https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-72ca9008b342e1caa5acc1466dfa7f20R37

packages/terra-clinical-result/src/common/utils.js
BenBoersma: Done https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-72ca9008b342e1caa5acc1466dfa7f20R48

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Removed all the unused imports https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-13926e711e02256769a4612c0d911268R13

packages/terra-clinical-result/CHANGELOG.md
BenBoersma: Fixed https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-534a7c4c0b2429787b66bb7aac9cade0R8

packages/terra-clinical-result/src/common/utils.js
BenBoersma: Renamed https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-72ca9008b342e1caa5acc1466dfa7f20R33

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Made _BloodPressureDisplay component https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-db5d557fed702f3b0bcac70aa9a018eaR9

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Fixed https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-13926e711e02256769a4612c0d911268R154

packages/terra-clinical-result/src/common/utils.js
BenBoersma: Made https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-8b312db510259892da0886d2f909d49eR89

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Welp ignore my paragraph response, turns out we can just do a ! statement
https://github.com/cerner/terra-clinical/pull/648/commits/bcd9623974696cc69c750cf77514f41a424ac509#diff-db5d557fed702f3b0bcac70aa9a018eaR39

packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
```
yuderekyu, line 1: we had a discussion in the past about removing underscored components and explicitly creating a private directory to contain private functionality. i can't find the discussion, but we've followed through in terra-application https://github.com/cerner/terra-application/tree/master/src/application-base/private

it doesn't have to be in this pull request, but we should plan on following that pattern in the future

packages/terra-clinical-result/src/common/observation/_Observation.jsx
yuderekyu, line 68: would optional chainging fulfill this check?
```suggestion
  const isValidValue = result?.value;
```

packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
```diff
@@ -0,0 +1,66 @@
...
+  type,
+  diastolicUnit,
+}) => {
+  if (!result) {
+    return <ResultError key={`Error-${type}-${id}`} />;
+  }
+  if (result.noData) {
+    return <NoData key={`NoData-${type}-${id}`} />;
+  }
+  const resultDisplay = (
```
yuderekyu, line 45: not a huge deal, but we can eliminate this intermediate variable and directly return
```suggestion
  return (
```

packages/terra-clinical-result/src/common/observation/_Observation.jsx
BenBoersma: I believe it does, so fixed it here https://github.com/cerner/terra-clinical/pull/648/commits/231120829ac48885ce4397bbe64fff91af5fefd8

packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
BenBoersma: Fixed https://github.com/cerner/terra-clinical/pull/648/commits/231120829ac48885ce4397bbe64fff91af5fefd8

packages/terra-clinical-result/CHANGELOG.md
ryanthemanuel: Is it a hefty list?  Or just more descriptive than we usually are :)

package.json
```diff
@@ -47,7 +47,8 @@
           "import/no-duplicates": "off"
         }
       }
-    ]
+    ],
+    "parser": "babel-eslint"
```
emilyrohrbough, line 51: Can we log an issue to eslint-config-terra to considering pulling this in to the default config?

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
emilyrohrbough: any update? or did I miss the changes?

package.json
yuderekyu: agreed, this is a great candidate for our default config. optional chaining for everyone :)

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Sorry addressed here, https://github.com/cerner/terra-clinical/pull/648/commits/415228472aa9b535d58726d4a69c09986fa23e4d#diff-13926e711e02256769a4612c0d911268R125
Discussed with Neil, we should just return a singular `--` if there is neither provided and not nothing like we currently do. Also added it to the no data test.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -120,145 +120,75 @@ const ClinicalResultBloodPressure = (props) => {
...
     );
   }
 
+  const decoratedResultClassnames = cx([
+    'decorated-result-display',
+    { truncated: isTruncated },
+    { 'status-in-error': systolicResult.statusInError || diastolicResult.statusInError },
+  ]);
+
+  clinicalResultBloodPressureDisplay = (
```
yuderekyu, line 178: can we move variable instantiation here? otherwise it's all the way at the top
```suggestion
  const clinicalResultBloodPressureDisplay = (
```

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -120,145 +120,75 @@ const ClinicalResultBloodPressure = (props) => {
   let clinicalResultBloodPressureDisplay = null;
```
yuderekyu, line 120: delete

packages/terra-clinical-result/src/common/utils.js
```diff
@@ -35,17 +71,10 @@ const checkTypeNumeric = (resultData) => {
...
 const ConditionalWrapper = ({ condition, wrapper, children }) => (condition ? wrapper(children) : children);
 
-const ClinicalResultUtils = {
-  isEmpty,
-  checkIsStatusInError,
-  checkTypeNumeric,
-  ConditionalWrapper,
-};
-
-export default ClinicalResultUtils;
```
tbiethman, line 45: Removing this default export is technically non-passive unless we've explicitly marked this file as private. We have taken to creating  a `/private` directory in other repos to explicitly mark files as not for consumer use and to give us leeway to make changes like this.

I don't anticipate that we have any direct consumer usage of this file, so I bring this up more as an FYI than a call for changes. But it's important for us to think about what APIs we expose from our packages, because often it's more than just the component definition.

package.json
ryanthemanuel: Added: https://github.com/cerner/terra-toolkit/issues/415

packages/terra-clinical-result/CHANGELOG.md
ryanthemanuel: I'm fine with leaving it as is.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Fixed here https://github.com/cerner/terra-clinical/pull/648/commits/c64377629f953ac331135eeef3730d8dc9ccb321

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
BenBoersma: Deleted here https://github.com/cerner/terra-clinical/pull/648/commits/c64377629f953ac331135eeef3730d8dc9ccb321

packages/terra-clinical-result/src/common/utils.js
BenBoersma: Since it is a MVB, I have reverted the changes and will put the change in another PR https://github.com/cerner/terra-clinical/pull/648/commits/b866ddd2f17961f96d98cd25553d8f0254d61d11

PR #679, opened 2020-07-13
Improve build times

mjhenkes:
### Summary

Similarly to [terra-core#3072](https://github.com/cerner/terra-core/pull/3072) we're running tests directly on travis instead of a docker container and re-organizing the steps to be more time efficient.  

Closes #

### Deployment Link

https://terra-clinical-deployed-pr-#.herokuapp.com/

### Additional Details

Here's our baseline:
Screen Shot 2020-07-13 at 12 34 06 PM

This PR:
Screen Shot 2020-07-13 at 1 31 07 PM

@cerner/terra

### DIFF
--- a/wdio.conf.js
+++ b/wdio.conf.js
@@ -8,4 +8,10 @@ if (process.env.npm_package_name !== 'terra-clinical') {
   wdioConfig.specs = [path.join(__dirname, 'packages', process.env.npm_package_name, 'tests', 'wdio', '**', '*-spec.js')];
 }
 
+const travis = process.env.TRAVIS;
+
+if (travis) {
+  wdioConfig.host = 'localhost';
+}
+
 exports.config = wdioConfig;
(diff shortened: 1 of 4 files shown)


wdio.conf.js
```diff
@@ -8,4 +8,10 @@ if (process.env.npm_package_name !== 'terra-clinical') {
   wdioConfig.specs = [path.join(__dirname, 'packages', process.env.npm_package_name, 'tests', 'wdio', '**', '*-spec.js')];
 }
 
+const travis = process.env.TRAVIS;
+
+if (travis) {
+  wdioConfig.host = 'localhost';
```
emilyrohrbough, line 14: Does this configuration when using the selenium -docker resolve the error you were seeing and the need to use the standalone-chrome-travis service?

mjhenkes: This lets us call the stand Alone chrome docker instance I spin up per executor. So we still need it and aren’t using the selenium docker service. Seems like it was having trouble promptly starting.

PR #773, opened 2021-08-24
Create stale.yml to auto close old issues

ryanthemanuel:
### Summary

Create stale.yml to auto close old issues


.github/stale.yml
```diff
@@ -0,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
```
ryanthemanuel, line 2: Do we want 2 months?  6?

.github/stale.yml
```diff
@@ -0,0 +1,16 @@
...
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - UX Audit
```
ryanthemanuel, line 7: Any other labels to exclude?

.github/stale.yml
```diff
@@ -0,0 +1,16 @@
...
+# Label to use when marking an issue as stale
+staleLabel: wontfix
```
ryanthemanuel, line 9: Any other label we want to use?

.github/stale.yml
```diff
@@ -0,0 +1,16 @@
...
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
```
ryanthemanuel, line 11: This is the comment we will use when marking as stale.

.github/stale.yml
mjhenkes: two works for me. lol this will probably mean that the issues get closed before the PR gets merged.

.github/stale.yml
mjhenkes: wontfix isn't the right sentiment, that implies that we acknowledge it's a problem. How about 'inactive'?

.github/stale.yml
ryanthemanuel: Maybe the Up Next labels?  To avoid sort of active work?

.github/stale.yml
ryanthemanuel:
```suggestion
staleLabel: inactive
```

.github/stale.yml
```diff
@@ -0,0 +1,16 @@
...
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
```
mjhenkes, line 14:
```suggestion
  This issue has been automatically marked as inactive because it has not had
  recent activity. It will be closed in seven days if no further activity occurs. Thank you
  for your contributions.
```

.github/stale.yml
mjhenkes: I actually kind of like the default one, with a modification.

.github/stale.yml
mjhenkes: that makes sense

benbcai: What about "bug" but i guess we usually don't ignore bug issues for 60 days.
What about "Priority: High"? We don't use this much but again we shouldn't ignore high priority issue for 60 days.

.github/stale.yml
mraplumb: I agree with Matt's changes.  I don't think we really need to state anything else.

tbiethman (approved):

.github/stale.yml
ryanthemanuel, line 7:
```suggestion
  - UX Audit
  - Up Next
  - bug
  - Priority: High
```

.github/stale.yml
```diff
@@ -0,0 +1,19 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - UX Audit
+  - Up Next
+  - bug
+  - Priority: High
```
ryanthemanuel, line 10:
```suggestion
  - "Priority: High"
```

mraplumb (approved):

mjhenkes (approved):

PR #816, opened 2022-10-13
[terra-clinical-data-grid] Fixed staircasing issue on IE due to precision loss.

saket2403:
### Summary

Rounded off total column width to the next integer to fix staircasing issue on IE caused due to precision loss when using float values for column  width.

Closes: UXPLATFORM-7755

### Deployment Link

https://terra-clinic-data-grid--7sdx5d.herokuapp.com/components/terra-clinical-data-grid/clinical-data-grid/clinical-data-grid

### Additional Details

@cerner/terra


supreethmr:
Can you update one of the test / example to have float values as width so that we will have view to test on IE when it is required in future.

saket2403:
> Can you update one of the test / example to have float values as width so that we will have view to test on IE when it is required in future.

Done f65922e

ShettyAkarsh (approved):

PR #821, opened 2023-03-03
Migrate to utilize GitHub Actions

poloka:
Migrate to utilize GitHub Actions

### Notes
Current Travis CI is failing in the main builds so this PR is just to move from failing in Travis to failing in GitHub.


sdadn (approved):
+1 since test failures are caused by other changes

PR #832, opened 2023-04-10
[terra-clinical-data-grid] Adding roles for screen reader usage.

BenBoersma:
### Summary

- Added roles to define data-grid so that screen readers have a better time reading.
- Column headers defined and the first cell in a row that is not the selection cell defined as row header.
- Added Label and Description ref props to allow consumers to define label and descriptions for data-grid.

Closes #
UXPLATFORM-8611

### Additional Details

@cerner/terra


packages/terra-clinical-data-grid/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Added
+  * Defined roles for better screen reader experience
+  * New labelRef and descriptionRef props for allowing addition label and description of data-grid
```
sdadn, line 7: nit:
```suggestion
  * New `labelRef` and `descriptionRef` props for allowing addition label and description of data-grid
```

Also, did you mean, _additional_?

packages/terra-clinical-data-grid/src/DataGrid.module.scss
```diff
@@ -122,4 +122,14 @@
       flex: 1 1 auto;
     }
   }
+
+  .hidden-info {
```
smason0, line 126: I wasn't sure on the naming of this className for the hidden label/description. Open to suggestions

packages/terra-clinical-data-grid/src/terra-dev-site/test/clinical-data-grid/LabelExternalDataGrid.test.jsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
```
smason0, line 1: This test file and the next one were being used for local devsite testing and may not need to be merged to main

packages/terra-clinical-data-grid/CHANGELOG.md
BenBoersma: Committed your suggestion and added this commit to have a better description of the changes https://github.com/cerner/terra-clinical/pull/832/commits/a7041d3c5f960636f0eec9355849a7a681e500b9

packages/terra-clinical-data-grid/src/DataGrid.module.scss
trandrew1023: Would Terra VisuallyHiddenText be sufficient for this scenario? We also added something similar to slide panel https://engineering.cerner.com/terra-ui/components/cerner-terra-core-docs/visually-hidden-text/about

packages/terra-clinical-data-grid/src/DataGrid.jsx
```diff
@@ -1000,10 +1024,12 @@ class DataGrid extends React.Component {
       ariaStyles['aria-hidden'] = true;
     } else if (isPinned) {
       ariaStyles.id = `${id}-Pinned-Row-${row.id}-Section-${section.id}`;
-      ariaStyles['aria-owns'] = `${id}-Overflow-Row-${row.id}-Section-${section.id}`;
+      // ariaStyles['aria-owns'] = `${id}-Overflow-Row-${row.id}-Section-${section.id}`;
```
trandrew1023, line 1027: Is this still needed?

packages/terra-clinical-data-grid/src/DataGrid.module.scss
smason0: I will try using that. That seems like what we'd want to use here if it means we can avoid adding custom css

packages/terra-clinical-data-grid/src/DataGrid.jsx
BenBoersma: If we don't comment this line out, screen readers cannout read the overflow columns. So it seems that this is actively messing with the intention of these codings. I meant to leave a comment above this one explaining why we commented it out.

packages/terra-clinical-data-grid/src/DataGrid.jsx
```diff
@@ -296,6 +314,9 @@ class DataGrid extends React.Component {
     document.addEventListener('keydown', this.handleKeyDown);
     document.addEventListener('keyup', this.handleKeyUp);
 
+    this.getLabelText();
+    this.getDescriptionText();
```
eawww, line 318: Might be an edge case, but if the element either of these refers to changes, will the accessible text update as well?

smason0: The way it is implemented currently it shouldn't update based on those props values changing because it is only setting state when this component mounts. To do that, we'd probably want to add these calls to the [`componentDidUpdate`](https://react.dev/reference/react/Component#componentdidupdate) lifecycle method.

smason0: UpdateLabelDesc

Was able to get this to update the hidden text (see spans in gif) using componentDidUpdate. Will be in my next commit

sdadn (approved):

packages/terra-clinical-data-grid/src/DataGrid.module.scss
smason0: Addressed here: https://github.com/cerner/terra-clinical/commit/b2683006e880a6d9b60d7939bd96273b829a3f80

eawww (approved):
Thanks for the great work on this!

PR #840, opened 2023-05-03
[terra-clinical-label-value-view] Changed control to be clearly described by the screen readers

MadalinaStelea:
### Summary

**What was changed:**
Changed terra-clinical-label-value-view control to be a definition list instead of a div.

**Why it was changed:**
The change was required for the screen readers to clearly describe the terra-clinical-label-value-view control as a definition list, with the label as a term and the value as description of the term..

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
jest_results
example_1
example_2 1
example_2 2
example_2 3
example_3

- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-8964 

---

@cerner/terra


sdadn (changes requested):
There should be jest snapshot updates with this change as well.

MadalinaStelea:
> There should be jest snapshot updates with this change as well.

Looks like I missed the jest snapshot yesterday. I've added it now, but the jest is still failing in the PR. I have also added a screenshot with jest tests passing on my local. Any idea what I am missing and why jest are passing on my local but are failing in here? On my local they are passing with and without -- -u at the end of the command.

MadalinaStelea:
> > > There should be jest snapshot updates with this change as well.
> > 
> > 
> > @sdadn Looks like I missed the jest snapshot yesterday. I've added it now, but the jest is still failing in the PR. I have also added a screenshot with jest tests passing on my local. Any idea what I am missing and why jest are passing on my local but are failing in here? On my local they are passing with and without -- -u at the end of the command.
> 
> Looks like there are some dependency updates in other components that require snapshot updates. I'll go ahead an update those in a separate PR.

Thank you, Saad!

sdadn:
Merged #842 which should fix the Jest build failures.

PR #843, opened 2023-05-04
[terra-clinical-item-display] Add hidden text meaning to strikethrough style

smason0:
### Summary

**What was changed:**
Added new `meaning` prop to allow consumers to assign a meaning to the ItemDisplay text, for screen readers to use. The strikethrough style will have a default meaning of 'deletion' in English, to simulate the behavior of the semantic `<del>` HTML element.

**Why it was changed:**
This change allows users of screen readers to be aware of text having strikethrough styling, to meet a11y requirements. It also allows consumers to define a meaning for any text, regardless of the style.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [x] Other (please describe below)
- [ ] No tests are needed

Tested with JAWS in Edge browser and VoiceOver in Safari, Chrome, and Firefox.

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-6840 

---

@cerner/terra

### DIFF
--- a/packages/terra-clinical-item-collection/tests/jest/__snapshots__/Item.test.jsx.snap
+++ b/packages/terra-clinical-item-collection/tests/jest/__snapshots__/Item.test.jsx.snap
@@ -11,12 +11,8 @@ exports[`List View Tests should pass customProps to the Item 1`] = `
     accessoryAlignment="alignCenter"
     displays={
       Array [
-        <ItemDisplay
-          iconAlignment="center"
-          isDisabled={false}
-          isTruncated={false}
+        <InjectIntl(ItemDisplay)
           text="Display 1"
-          textStyle="primary"
         />,
       ]
     }
@@ -67,26 +63,14 @@ exports[`List View Tests should render a Item with a list item chevron 1`] = `
     }
     displays={
       Array [
-        <ItemDisplay
-          iconAlignment="center"
-          isDisabled={false}
-          isTruncated={false}
+        <InjectIntl(ItemDisplay)
           text="Display 1"
-          textStyle="primary"
         />,
-        <ItemDisplay
-          iconAlignment="center"
-          isDisabled={false}
-          isTruncated={false}
+        <InjectIntl(ItemDisplay)
           text="Display 2"
-          textStyle="primary"
         />,
-        <ItemDisplay
-          iconAlignment="center"
-          isDisabled={false}
-          isTruncated={false}
+        <InjectIntl(ItemDisplay)
           text="Display 3"
-          textStyle="primary"
         />,
       ]
     }
@@ -118,12 +102,8 @@ exports[`List View Tests should render a Item with a single children 1`] = `
     accessoryAlignment="alignCenter"
     displays={
       Array [
-        <ItemDisplay
-          iconAlignment="center"
-          isDisabled={false}
-          isTruncated={false}
+        <InjectIntl(ItemDisplay)
           text="Display 1"
-          textStyle="primary"
         />,
       ]
     }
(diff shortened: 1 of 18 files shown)


packages/terra-clinical-item-display/src/ItemDisplay.jsx
```diff
@@ -43,6 +44,16 @@ const propTypes = {
    * One of `'center'`, `'top'`, `'inline'`.
    */
   iconAlignment: PropTypes.oneOf(['center', 'top', 'inline']),
+  /**
+   * The meaning of the text based on styling, used by screen readers.
+   * Changing `meaning` will not visually change the style of the content.
+   */
+  meaning: PropTypes.string,
```
sdadn, line 51: `meaning` is a bit vague. Can we update this to something like `textMeaning` (or any other suitable alternative)?

packages/terra-clinical-item-display/tests/jest/ItemDisplay.test.jsx
```diff
@@ -1,79 +1,109 @@
...
 it('should render a default component', () => {
-  const itemDisplay = shallow(<Display />);
+  const itemDisplay = shallowWithIntl(<Display />).shallow();
   expect(itemDisplay).toMatchSnapshot();
 });
 
 it('should render with text', () => {
-  const itemDisplay = shallow(<Display text="Display" />);
+  const itemDisplay = shallowWithIntl(<Display text="Display" />).shallow();
+  expect(itemDisplay.find('.text').prop('children')).toEqual('Display');
```
sdadn, line 14: +1 on adding these test cases. Thanks for noticing and adding these!

mjpalazzo:
Validated with JAWS and VO. Screen readers announce the strikethrough variant. Nice work!

smason0:
I can also add a doc site example of this prop's usage.

packages/terra-clinical-item-display/src/ItemDisplay.jsx
```diff
@@ -43,6 +44,16 @@ const propTypes = {
    * One of `'center'`, `'top'`, `'inline'`.
    */
   iconAlignment: PropTypes.oneOf(['center', 'top', 'inline']),
+  /**
```
smason0, line 47: I can also add the "Accessibility" badge here, that will show in the doc site props

packages/terra-clinical-item-display/src/ItemDisplay.jsx
```diff
@@ -43,6 +44,16 @@ const propTypes = {
...
+   * The meaning of the text based on styling, used by screen readers.
+   * Changing `meaning` will not visually change the style of the content.
+   */
```
adoroshk, line 50: Unless it was omitted in purpose, it might be worth adding a line here, something like: `For strike through text style only, the default value is 'deletion'`. Currently the Display Props Table says that `meaning` has no default value (which is true for all cases, but strike through text).

packages/terra-clinical-item-display/src/ItemDisplay.jsx
adoroshk: > `meaning` is a bit vague. Can we update this to something like `textMeaning` (or any other suitable alternative)?

Good catch, I would even expand that to read `textStyleMeaning`, as my understanding is we are adding this prop to explain the visual styling of the text rather than text itself to the assistive technology

packages/terra-clinical-item-display/src/ItemDisplay.jsx
smason0: I can add a note about strikethrough to the prop description.

packages/terra-clinical-item-collection/tests/jest/__snapshots__/Item.test.jsx.snap
```diff
@@ -11,12 +11,8 @@ exports[`List View Tests should pass customProps to the Item 1`] = `
...
-        <ItemDisplay
-          iconAlignment="center"
-          isDisabled={false}
-          isTruncated={false}
+        <InjectIntl(ItemDisplay)
           text="Display 1"
-          textStyle="primary"
         />,
       ]
     }
```
sdadn, line 18: These tests will need updating as these snapshots are now too high level due to the addition of internationalization. However, it is now outside the scope of this PR so we can make a follow-up Jira for it.

smason0: Logged UXPLATFORM-9008

sdadn (approved):

trandrew1023 (approved):

adoroshk (approved):
Tested locally, works as expected. The tests fixes being a separate JIRA works for me.

PR #845, opened 2023-05-05
[terra-clinical-item-display] Deprecating isDisabled prop

sdadn:
### Summary

Deprecated `isDisabled` prop as this feature does not meet accessibility guidelines. 

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-8934 

---

@cerner/terra

### DIFF
--- a/packages/terra-clinical-item-display/src/ItemDisplay.jsx
+++ b/packages/terra-clinical-item-display/src/ItemDisplay.jsx
@@ -29,8 +29,9 @@ const propTypes = {
    * Whether or not the text should be truncated.
    */
   isTruncated: PropTypes.bool,
+  //  TODO: remove isDisabled in the next major release.
   /**
-   * Whether or not the display is disabled.
+   * (Deprecated) Whether or not the display is disabled.
    */
   isDisabled: PropTypes.bool,
   /**
@@ -81,6 +82,11 @@ const ItemDisplay = ({
     { [`${textStyle}`]: textStyle === TextStyles.STRONG },
     { 'strike-through': textStyle === TextStyles.STRIKETHROUGH },
   ]);
+
+// TODO: remove this warning in the next major release.
+  if(isDisabled){
+    console.warn('The isDisabled prop does not meet a11y standards and should not be used. It will be removed in the next major release.');
+  }
 
   let displayIcon;
   if (icon) {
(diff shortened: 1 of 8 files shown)


sycombs (approved):

eawww (commented):
Could you also remove the isDisabled example from the doc site?

sdadn:
> Could you also remove the isDisabled example from the doc site?

Since the prop is still usable, would it make sense to keep the example and remove it when the prop is fully removed?

packages/terra-clinical-item-display/src/ItemDisplay.jsx
```diff
@@ -82,6 +83,11 @@ const ItemDisplay = ({
     { 'strike-through': textStyle === TextStyles.STRIKETHROUGH },
   ]);
 
+  // TODO: remove this warning in the next major release.
+  if (isDisabled) {
```
smason0, line 87: I see it's just a warning, but should we disable the no-console linter rule for this line?

packages/terra-clinical-item-display/src/ItemDisplay.jsx
```diff
@@ -29,8 +29,9 @@ const propTypes = {
    * Whether or not the text should be truncated.
    */
   isTruncated: PropTypes.bool,
+  //  TODO: remove isDisabled in the next major release.
```
smason0, line 32: Is there a Jira logged to go back and address these TODOs?

sdadn:
> > Since the prop is still usable, would it make sense to keep the example and remove it when the prop is fully removed?
> 
> @sdadn I wouldn't say so. The example is to instruct consumers on consuming it. Nobody should be implementing what the example shows now that it's deprecated. I can't think of a benefit it provides and may actually encourage new use of it.

Since this doesn't affect passivity, I went ahead and removed the examples here:
d709832991c616adbf63cab791d2a27c91641c72

packages/terra-clinical-item-display/src/ItemDisplay.jsx
sdadn: Disabled here: 51180274acf452a3d57b7018e52ed217271efdfe

eawww (approved):

smason0 (approved):

PR #854, opened 2023-05-18
[terra-clinical-item-view] Updating jests to properly consume item-display

BenBoersma:
### Summary
 Updating consumption of terra-clinical-item-display in jest tests since with recent INTL changes it became to high level to just shallow render.

**What was changed:**
Jest tests were changed to better consume and render `item-display` in snapshots while also updating tests to test specific things to not only rely on Jest snapshots.

**Why it was changed:**
`terra-clinical-item-display` was recently update to use `injectintl` function which was causing it to only render high level shallows of itself in Jest. Also tests were dependent on just jest snapshots and no other criteria.

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9008 

---

@cerner/terra

### DIFF
--- a/packages/terra-clinical-item-view/CHANGELOG.md
+++ b/packages/terra-clinical-item-view/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## Unreleased
 
-* Changed
-  * Updated Jest tests
-
 ## 4.7.0 - (May 15, 2023)
 
 * Changed
--- a/packages/terra-clinical-item-view/CHANGELOG.md
+++ b/packages/terra-clinical-item-view/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Updated Jest tests
+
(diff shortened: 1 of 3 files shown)


packages/terra-clinical-item-view/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Updated Jest tests
+
```
sdadn, line 7: This is not needed as this change has no consumer value and it will essentially trigger a duplicate release.
```suggestion
```

sdadn (approved):
Left a comment on the CHANGELOG but otherwise looks good!

sycombs (approved):

PR #859, opened 2023-05-25
[terra-clinical-result] Read out 'Value Uncharted as In Error' when s…

chenyixuan625:
…tatus is entered-in-error

### Summary

**What was changed:**
Creating an additional span that is visually hidden but has the 'Value Uncharted as In Error' text for screen reader usage.

**Why it was changed:**
The accessibility tooling needs to read out the correct string for the FlowsheetResultCell with the "entered-in-error" status.

### Testing

This change was tested using: 
Jest, WDIO tests and Screen reader testing
Finding a FlowsheetResultCell with the status "entered-in-error", and ensure the accessibility tooling reads out "Value Uncharted as In Error".

https://github.com/cerner/terra-clinical/assets/44185348/a69e7d07-bf42-4efb-b6e6-6b132c867da4

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9065 

---

@cerner/terra


packages/terra-clinical-result/src/common/other/_EnteredInError.jsx
```diff
@@ -36,9 +37,10 @@ const EnteredInError = (props) => {
       {...customProps}
       className={templateClassnames}
     >
-      <span aria-label={intl.formatMessage({ id: 'Terra.clinicalResult.statusInErrorAria' })}>
+      <span aria-hidden="true">
```
RayGunY, line 40: Do we want to hide this? The implementation before implies that they wanted it to read the Terra.clinicalResult.statusInError message but also then the Terra.clinicalResult.statusInErrorAria message. 
It would read as "In error. Value uncharted as In Error." which I think makes sense - I can see the argument that maybe it's repetitive but I'm not sure if that's intentional or if we can skip it. Definitely need to double check this with Scott. 

The reason why I'm using aria-hidden for my jira is because '--' is never read out by the screen reader so it makes sense to hide that from the voice over.

sdadn (commented):
This looks good from an engineering standpoint. However I'm a bit unclear about the screenreader changes so I'm waiting on accessibility approval before I can approve & merge this PR.

CONTRIBUTORS.md
```diff
@@ -40,6 +40,7 @@ Cerner Corporation
 - Dianna McGinn [@DMcginn]
 - Ashish kumbhare [@ashishkcerner]
 - Razvan Luca [@razvanluca]
+- Yixuan Chen [@chenyixuan625]
```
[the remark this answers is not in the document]
chenyixuan625: Added: https://github.com/cerner/terra-clinical/pull/859/commits/e773fef2dd5b6277ef86a0ab41476299e62e56c8

packages/terra-clinical-result/src/common/other/_EnteredInError.jsx
chenyixuan625: I checked with Scott this morning that only "Value uncharted as In Error" needs to be read out by the screen reader.
cc @scottwilmarth

CONTRIBUTORS.md
```diff
@@ -84,4 +85,5 @@ Community
 [@saket2403]: https://github.com/saket2403
 [@DMcginn]: https://github.com/DMcginn
 [@ashishkcerner]: https://github.com/ashishkcerner
-[@razvanluca] https://github.com/razvanluca
\ No newline at end of file
+[@razvanluca] https://github.com/razvanluca
+[@chenyixuan625] https://github.com/chenyixuan625
```
vinaybhargavar, line 89: nit: We will need to add a `:` after our username.

chenyixuan625:
@chrismichalewicz Would you mind reviewing it when you get a second? Thanks!

chrismichalewicz:
> @chrismichalewicz Would you mind reviewing it when you get a second? Thanks!

@chenyixuan625 Tested this with JAWS and it looks good. Updating the A11y label. Thanks!

kolkheang (approved):

sdadn (approved):

sycombs (approved):

PR #860, opened 2023-05-25
[terra-clinical-result] Add 'No Result' hidden text

RayGunY:
### Summary

**What was changed:**

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra


vinaybhargavar:
nit: We will also need to update the contributors' file.

packages/terra-clinical-result/src/common/other/_KnownNoData.jsx
```diff
@@ -35,9 +36,11 @@ const KnownNoData = (props) => {
     <span
       {...customProps}
       className={templateClassnames}
-      aria-label={intl.formatMessage({ id: 'Terra.clinicalResult.resultNoData' })}
     >
-      --
+      <span aria-hidden>
```
chenyixuan625, line 40: Just curious, is there any difference between `<span aria-hidden>` and `<span aria-hidden="true">` from the accessibility tooling perspective? I notice the aria-hidden is usually set with a value in other places.

RayGunY: I was given an error message from the linter when I tried explicitly setting it to true that said "Value must be omitted for boolean attributes". So basically it looks like it just defaults to true so we don't have to set it.

RayGunY:
> nit: We will also need to update the contributors' file.

Updated here: [d6b6325](https://github.com/cerner/terra-clinical/pull/860/commits/d6b6325b2267a65134c375d0ca5539f3f23f0de1)

packages/terra-clinical-result/tests/jest/common/other/KnownNoData.test.jsx
```diff
@@ -6,6 +6,7 @@ import KnownNoData from '../../../../src/common/other/_KnownNoData';
 describe('KnownNoData', () => {
   it('should render a default KnownNoData', () => {
     const result = shallowWithIntl(<KnownNoData />).dive();
+    expect(result.find('VisuallyHiddenText').at(0).prop('text')).toEqual('Terra.clinicalResult.resultNoData');
```
sdadn, line 9: +1 on the test case

packages/terra-clinical-result/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Added
   * Added screen-reader support for clinical-result icons. (Requires Jest test updates on consuming applications)
+  * Added screen-reader support for alternative text for when no results are returned
```
sdadn, line 7:
```suggestion
  * Added screen-reader support for alternative text for when no results are returned.
```

sdadn (commented):
This looks good from an engineering standpoint. However I'm a bit unclear about the screenreader changes so I'm waiting on accessibility approval before I can approve & merge this PR.

packages/terra-clinical-result/CHANGELOG.md
RayGunY: Good catch, fixed: https://github.com/cerner/terra-clinical/pull/860/files#diff-2f347c01ccae568264be55304c9e1799548817ccb16246b00f7baa02940f2e05R7

packages/terra-clinical-result/src/common/other/_KnownNoData.jsx
RayGunY: Hey so I looked into this more this morning, I made a change to set it to true explicitly. aria-hidden isn't a boolean but a state that can have three different values. It defaults to 'undefined' which is working in my case for the span but is not reliable - anyways here's some more info: https://stackoverflow.com/questions/51255045/is-presence-of-aria-hidden-sufficient-or-is-value-set-to-true-required-aria-h

vinaybhargavar (approved):

chenyixuan625 (approved):

sycombs (approved):

PR #865, opened 2023-06-06
[terra-clinical-result] Update Result Time/Name header cells and FlowsheetResultCell to use proper HTML tag

RayGunY:
### Summary
FlowsheetResultCell, ResultNameHeaderCell, and ResultTimeHeaderCell are presumed to be used as part of a table in which they do not provide the proper context of a Cell or Header Cell for a table. 

**What was changed:**
We need to change the HTML tags from div to the proper corresponding tags for Header Cells and Cells. That would be _td_ for FlowsheetResultCell and _th_ for ResultNameHeaderCell and ResultTimeHeaderCell.

Also needed to add site tests for each in order to properly validate this change, this is because the 'tables' created on the site currently are not actual semantic html tables. That means the screenreader doesn't pick up on the new context of cell or header cell, just reads them as it has been previously. This is good in that changing the tags doesn't cause any visual changes to current implementation if you aren't using a table.

That being said, when using the semantic html table for certain flowsheet result cell entries I'm having issues with the css formatting - specifically when dealing with the spacing of certain icons or with the multiple result custom icon being too wide. Overall the cell itself is also much wider than seemingly needed - I think this and the icon issue are caused because of the actual content styling inside the ResultFlowsheetCell so I'm not sure yet what I might need to do. Looking into that now. 
Example of the styling issues:
Example-of-bad-css

Either way we still need to test it in a real table so I opted to use terra's existing [html table component](https://engineering.cerner.com/terra-ui/components/cerner-terra-core-docs/html-table/about) in order to build that out.

**Why it was changed:**
We need proper context for these components when they are used within a semantic table.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9063

---

@cerner/terra

Flowsheet cell test:

https://github.com/cerner/terra-clinical/assets/40574651/7bc953c8-64eb-40c1-9c77-b8bc74f9a8cf

https://github.com/cerner/terra-clinical/assets/40574651/adbcd741-115f-433e-936b-daa6243f4cd5

Header cell tests:

https://github.com/cerner/terra-clinical/assets/40574651/e762b670-0e39-4311-b53e-3a7af8bcf6d3

https://github.com/cerner/terra-clinical/assets/40574651/df0640f4-10bb-4623-b449-688dd398aa6a

### DIFF
--- a/packages/terra-clinical-result/CHANGELOG.md
+++ b/packages/terra-clinical-result/CHANGELOG.md
@@ -9,6 +9,9 @@
 * Fixed
   * Fixed a check related to Clinical Result strikethrough alt text for if a result unit exists or not.
 
+* Changed
+  * Changed FlowsheetResultCell, ResultNameHeaderCell and ResultTimeHeaderCell in clinical-result to use proper semantic html.
+
 ## 1.17.0 - (June 14, 2023)
 
 * Added
@@ -18,6 +21,7 @@
   * Added screen-reader support for clinical-result FlowsheetResultCell with multiple results.
   * Added screen-reader support for clinical-result ResultNameHeaderCell. Two new optional props are added for providing a full name of the result name and unit respectively.
   * Added screen-reader support for Blood Pressure with additional displays as a group. An additional prop is added for making the blood pressure results a group.
+
 
 ## 1.16.0 - (March 29, 2023)
 
--- a/packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
+++ b/packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
@@ -77,6 +77,7 @@ const createEndIcons = (hasCommentIcon, hasModifiedIcon, hasUnverifiedIcon, resu
   if (!hasCommentIcon && !hasModifiedIcon && !hasUnverifiedIcon) {
     return null;
   }
+
   let iconElements;
   if (hasUnverifiedIcon) {
     iconElements = <IconUnverified className={cx('icon-unverified')} a11yLabel={intl.formatMessage({ id: 'Terra.clinicalResult.resultUnverified' })} role="img" focusable="true" />;
@@ -88,6 +89,7 @@ const createEndIcons = (hasCommentIcon, hasModifiedIcon, hasUnverifiedIcon, resu
       </React.Fragment>
     );
   }
+
   return (
     <div key={(`EndAccessoryIcons-${resultKeyID}`)} className={cx('end-accessory-icons')}>
       <div className={cx('end-accessory-stack')}>
@@ -102,6 +104,7 @@ const createEndAdditionalResultsStack = (count, interpretationsArr, hasAccessory
   if (displayCount < 1) {
     return null;
   }
+
   let additionalResultInterpretationIndicator;
   let criticality;
   if ([
@@ -122,6 +125,7 @@ const createEndAdditionalResultsStack = (count, interpretationsArr, hasAccessory
   } else {
     criticality = 'normal';
   }
+
   const additionalResultClassNames = cx([
     'additional-end-display',
     { 'no-accessory-icons': !hasAccessoryIcons },
@@ -134,6 +138,7 @@ const createEndAdditionalResultsStack = (count, interpretationsArr, hasAccessory
   const text = (criticality === 'normal')
     ? intl.formatMessage({ id: 'Terra.clinicalResult.multipleNormalResults' }, { count })
     : intl.formatMessage({ id: 'Terra.clinicalResult.multipleResults' }, { count, criticality });
+
   return (
     <div key={(`AdditionalResultsDisplay-${resultKeyID}`)} className={additionalResultClassNames}>
       <div className={cx('additional-results-stack')} aria-hidden="true">
@@ -150,6 +155,7 @@ const createClinicalResultDisplay = (children, hasUnverifiedIcon, hasInterpretat
     'primary-display',
     { interpretation: hasInterpretationIcon && !hasUnverifiedIcon },
   ]);
+
   return (<div key={(`ClinicalResultDisplay-${resultKeyID}`)} className={primaryResultClassnames} ref={containerDivRef}>{children}</div>);
 };
 
@@ -162,7 +168,9 @@ const createStandardResultDisplay = (resultDataItem, resultAttributes, hideUnit,
   } else {
     resultsInnerDisplay = <ClinicalResult key={(`ClinicalResult-${resultKeyID}`)} {...resultDataItem} hideUnit={hideUnit} isTruncated isUnverified={resultAttributes.unverified} hideAccessoryDisplays />;
   }
+
   const clinicalResultDisplay = createClinicalResultDisplay(resultsInnerDisplay, resultAttributes.unverified, resultAttributes.interpretationIcon, containerDivRef, resultKeyID);
+
   return clinicalResultDisplay;
 };
 
@@ -177,7 +185,9 @@ const createBloodPressureResultDisplay = (resultDataItem, resultAttributes, hide
   } else {
     resultsInnerDisplay = (<ClinicalResultBloodPressure key={(`ClinicalResultBloodPressure-${resultKeyID}`)} systolic={systolic} diastolic={diastolic} hideUnit={hideUnit} isTruncated hideAccessoryDisplays />);
   }
+
   const clinicalResultDisplay = createClinicalResultDisplay(resultsInnerDisplay, resultAttributes.unverified, resultAttributes.interpretationIcon, containerDivRef, resultKeyID);
+
   return clinicalResultDisplay;
 };
 
@@ -200,6 +210,7 @@ const setResultKeyID = (isBloodPressureResult, resultData) => {
       return resultData.eventId;
     }
   }
+
   return null;
 };
 
@@ -208,9 +219,11 @@ const checkIfSingleOrPairedResult = (resultDataItem) => {
   if (isSingleResult) {
     return { isSingleResult, isPairedResult: false };
   }
+
   const hasSystolicData = !isEmpty(resultDataItem.systolic) ? resultDataItem.systolic.result : false;
   const hasDiastolicData = !isEmpty(resultDataItem.diastolic) ? resultDataItem.diastolic.result : false;
   const isPairedResult = (hasSystolicData || hasDiastolicData) || false;
+
   return { isSingleResult, isPairedResult };
 };
 
@@ -236,6 +249,7 @@ const unpackResultAttributes = (resultDataItem) => {
   itemAttributes.comment = hasComment;
   itemAttributes.modified = isModified;
   itemAttributes.unverified = isUnverified;
+
   return itemAttributes;
 };
 
@@ -262,9 +276,11 @@ const unpackResultDataSet = (resultDataSet) => {
       (bpAttribute.systolic.unverified || bpAttribute.diastolic.unverified),
     );
   }
+
   const isfirstSingleResult = isSingleResult;
   const isfirstPairedResult = isPairedResult;
   const resultKeyID = setResultKeyID(isfirstPairedResult, firstResultData);
+
   return {
     isfirstSingleResult,
     isfirstPairedResult,
@@ -282,6 +298,7 @@ const createFlowsheetResultCellDisplay = (resultDataSet, hideUnit, numericOverfl
     firstResultData,
     resultKeyID,
   } = unpackResultDataSet(resultDataSet);
+  const hasAccessoryIcons = (firstResultAttributes.comment || firstResultAttributes.modified || firstResultAttributes.unverified);
   const compositeCell = [];
   if (!isfirstSingleResult && !isfirstPairedResult) {
     compositeCell.push(<ResultError />);
@@ -292,6 +309,7 @@ const createFlowsheetResultCellDisplay = (resultDataSet, hideUnit, numericOverfl
     const firstResultDisplay = createBloodPressureResultDisplay(firstResultData, firstResultAttributes, hideUnit, resultKeyID, containerDivRef);
     compositeCell.push(firstResultDisplay);
   }
+
   const additionalResultCount = resultDataSet.length - 1;
   if (additionalResultCount > 0) {
     const additionalResultInterpretations = [];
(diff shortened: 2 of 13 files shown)


kolkheang:
@RayGunY - I know you're still working on this PR... but just as FYI - there are a few tests that failed
https://github.com/cerner/terra-clinical/actions/runs/5191908268/jobs/9361357838?pr=865

RayGunY:
I've added updated recordings and the formatting of everything should be looking correct
@kolkheang @chenyixuan625 this should be good to review now

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
```diff
@@ -303,11 +303,33 @@ const createFlowsheetResultCellDisplay = (resultDataSet, hideUnit, numericOverfl
...
     compositeCell.push(additionalResultsStackDisplay);
   }
   const endAccessoryIcons = createEndIcons(firstResultAttributes.comment, firstResultAttributes.modified, firstResultAttributes.unverified, resultKeyID);
-  compositeCell.push(endAccessoryIcons);
+  const endAccessoryIconsDisplay = (
+    <div key="EndDisplay-Icons" className={cx('end-display')}>
+      {endAccessoryIcons}
+    </div>
+  );
+  compositeCell.push(endAccessoryIconsDisplay);
```
chenyixuan625, line 332: Are these changes specifically for the fix of additional results squares and end accessory icons? I pulled the branch to my local and run the site. Looks like squares and icons are showing up on the right side. Is this where we are right now?
Screenshot 2023-06-08 at 3 40 28 PM

RayGunY: This is a little complicated but here was my thought process:

When trying to format for a semantic table the spacing was off (you can see in the screenshot in the top description). After making changes to get them all on the same line those squares and icons still needed to be moved to the right, so I tried to use [float](https://www.w3schools.com/cssref/pr_class_float.php) on the them.

The way the logic in this file worked before was if there were additional results then the squares were created and pushed first, then if there were icons those were created and pushed. So visually if both were present then both would be on the right hand side of the cell with the squares coming first and the icons to the right of the squares.

The tricky part is when trying to use float it picks up whichever element was added first (squares) and moves it as far right as it can go, then the next one is added (icons) and is moved right until it hits the first element. So basically they were flipped in the wrong order, icons on the left and squares on the right. Saw some suggestions to just swap the order of the elements but because the rest of the UI currently uses non-semantic elements like divs - if we try to swap them it messes up that non-semantic formatting. 

My solution was to wrap the squares and the icons in a parent container so I could just float the entire container to the right while still maintaining the proper order of the squares and icons. I checked the rest of the tests and site examples and everything still looks the same, no wdio errors so the non-semantic formatting hasn't changed.

RayGunY: Also I'm using the parent container around the elements even if they are alone. This is so we don't have to worry about floating those individually when the other isn't there, they'll just always be in the container to the right. Otherwise when we only have the squares or only have the icons we'll continue to have the issue where they are on the left with the text.

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
```diff
@@ -303,11 +303,33 @@ const createFlowsheetResultCellDisplay = (resultDataSet, hideUnit, numericOverfl
     });
     const displayCount = additionalResultCount + 1;
     const hasAccessoryIcons = (firstResultAttributes.comment || firstResultAttributes.modified || firstResultAttributes.unverified);
-    const additionalResultsStackDisplay = createEndAdditionalResultsStack(displayCount, additionalResultInterpretations, hasAccessoryIcons, resultKeyID);
+    const additionalResultsStack = createEndAdditionalResultsStack(displayCount, additionalResultInterpretations, hasAccessoryIcons, resultKeyID);
+
+    if (hasAccessoryIcons) {
```
kolkheang, line 308: Is this grouping needed? The code below adds additionalResultsStack and endAccessoryIcons separately. What scenario is grouping for both needed vs separate results stack and accessory icons?

kolkheang (commented):
Could we test (by resizing the browser) the flex wrap for when content is rendering in various form factors to see how the icons and cell are displayed?

packages/terra-clinical-result/src/terra-dev-site/test/clinical-result/result-name-header-cell/SemanticTableWithResultNameHeaderCell.test.jsx
```diff
@@ -0,0 +1,24 @@
...
+      </Row>
+      <Row>
+        <ResultNameHeaderCell resultName="BP" unit="mmHg" />
+      </Row>
+      <Row>
+        <ResultNameHeaderCell resultName="HR" unit="mmHg" />
+      </Row>
+    </Header>
+  </Table>
+);
```
chenyixuan625, line 22: Looking at this new ResultNameHeaderCell example, is it possible to keep the name and unit at the same level(like the example in the component site)?
Screenshot 2023-06-09 at 2 12 46 PM

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
RayGunY: I did it like this because when there is an end accessory icon and additional results they need to be grouped in the same parent container. This is so the order of end icons shows up correctly - we need to float the whole container to the right hand side instead of the additional results stack icon and accessory icons individually (when both are present). When it's just the additional results stack or just the accessory icons they still are put into the container so we don't have to worry in the css about floating those individually. It's a bit complicated so more info about my thought process can be found here: https://github.com/cerner/terra-clinical/pull/865#discussion_r1223572374

To break it down further: 

1. If there are additional results and accessory icons we create them, put those into a container together, push it to the composite cell and return.
2. If there are additional results but no accessory icons then it creates just the additional results, puts it into the container, pushes it, and returns. We don’t need to continue to the next bit of logic since we already checked if there were accessory icons.
3. If no additional results then we just check if there are accessory end icons, if there are then we create them, put them in the container, and push. I realized I needed to create a check for hasAccessoryIcons so that we don’t create an empty div if createEndIcons returns null. Added that logic here: https://github.com/cerner/terra-clinical/pull/865/commits/8f533dc140ddcfdd4115e688807f767be79e73d2

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
```diff
@@ -303,11 +303,33 @@ const createFlowsheetResultCellDisplay = (resultDataSet, hideUnit, numericOverfl
...
+      const endAccessoryIcons = createEndIcons(firstResultAttributes.comment, firstResultAttributes.modified, firstResultAttributes.unverified, resultKeyID);
```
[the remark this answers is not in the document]
RayGunY: I not sure if we'd want to - if there are accessory icons then they'd be handled here and we'll return. I don't think we want to create them outside of this if they don't exist. The logic around 326 only happens if we can't do this check here (aka when additional results don't exist).

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
```diff
@@ -303,11 +303,33 @@ const createFlowsheetResultCellDisplay = (resultDataSet, hideUnit, numericOverfl
...
+      const endDisplay = (
+        <div key="EndDisplay-AdditionalResultsAndIcons" className={cx('end-display')}>
+          {additionalResultsStack}
+          {endAccessoryIcons}
+        </div>
+      );
+      compositeCell.push(endDisplay);
+      return compositeCell;
+    }
```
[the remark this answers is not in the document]
RayGunY: It seemed like in these files they're not really using newlines to break the code up much already. So while I wanted to add new lines I didn't because it seemed to go against some standard they're using - what do you think?

packages/terra-clinical-result/src/terra-dev-site/test/clinical-result/result-name-header-cell/SemanticTableWithResultNameHeaderCell.test.jsx
RayGunY: Good catch! Updated the css here: [002e458](https://github.com/cerner/terra-clinical/pull/865/commits/002e458494d9942bd483eb5904dc2bb8a411138e)

The tests are still running but here's what it looks like now:
Screenshot 2023-06-12 at 11 26 20 AM

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
kolkheang: Thanks for the additional info. I think this info would be good to include as a code comment.

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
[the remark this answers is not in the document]
RayGunY: Alright, added some newlines throughout the file: [be3aea5](https://github.com/cerner/terra-clinical/pull/865/commits/be3aea592e99436de5829afb8e9a03bdeb84ccfd)

Generally tried to add newlines to create some separation for the return statements and after if statements

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
RayGunY: Added a comment, let me know if it's clear enough. Was trying to stay concise and just mention the main reason for that logic block being there: [b063110](https://github.com/cerner/terra-clinical/pull/865/commits/b063110b15c2ad1f1345416f80380b51063bfe75)

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
```diff
@@ -266,6 +282,7 @@ const createFlowsheetResultCellDisplay = (resultDataSet, hideUnit, numericOverfl
     firstResultData,
     resultKeyID,
   } = unpackResultDataSet(resultDataSet);
+  const hasAccessoryIcons = (firstResultAttributes.comment || firstResultAttributes.modified || firstResultAttributes.unverified);
```
chenyixuan625, line 285: Just curious, any reason to bump this line up here?

packages/terra-clinical-result/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
   * Added screen-reader support for clinical-result icons. (Requires Jest test updates on consuming applications)
   * Added screen-reader support for alternative text for when no results are returned.
   * Added screen-reader support for clinical-result FlowsheetResultCell EnteredInError.
+  * Changed cells and header cells in clinical-result to use proper semantic html
```
chenyixuan625, line 9: nit: Could we call out the FlowsheetResultCell, ResultNameHeaderCell and ResultTimeHeaderCell specifically?

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
RayGunY: It's being used in two different places so I pulled it out to re-use it:

1. https://github.com/cerner/terra-clinical/blob/b063110b15c2ad1f1345416f80380b51063bfe75/packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx#LL326C7-L326C7
2. https://github.com/cerner/terra-clinical/blob/b063110b15c2ad1f1345416f80380b51063bfe75/packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx#L353

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
```diff
@@ -301,13 +319,46 @@ const createFlowsheetResultCellDisplay = (resultDataSet, hideUnit, numericOverfl
         }
       }
     });
+
     const displayCount = additionalResultCount + 1;
-    const hasAccessoryIcons = (firstResultAttributes.comment || firstResultAttributes.modified || firstResultAttributes.unverified);
-    const additionalResultsStackDisplay = createEndAdditionalResultsStack(displayCount, additionalResultInterpretations, hasAccessoryIcons, resultKeyID);
+    const additionalResultsStack = createEndAdditionalResultsStack(displayCount, additionalResultInterpretations, hasAccessoryIcons, resultKeyID);
```
kolkheang, line 324: Can we add a new line here too? Between the const declaration and the comment.

kolkheang (approved):

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
RayGunY: Yup! https://github.com/cerner/terra-clinical/blob/1feea097b7fad242ae9e7d622d5ec465c11c3688/packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx#L329

kolkheang (approved):

brooksienoodlesoup (commented):
Tested with:
Chrome 108.0.5359.94 (Official Build) (arm64) + VO (Ventura 13.2.1)
Edge 114.0.1823.43 (Official build) (64-bit) + JAWS 2023.2303.144

*NOTE* - In order to test table headers or table data cells, I can only really test them with a screen reader when they are combined into a full fledged table that has both, like you set up with the Semantic Table with Flowsheet Result Cell.  

- Results for VO were as expected.
- Results for Jaws did not work as expected.  

### Jaws testing details
In Jaws, column headers were not announced by default. Jaws didn't switch into "Table Layer Mode" automatically when I navigated to the table.  I had to manually switch into "Table Layer Mode" by pressing Ins+Spacebar, then "T" on my PC keyboard.  Then, I was able to navigate the table and hear the proper announcement. ([HTML Tables with Jaws keyboard instructions](https://www.freedomscientific.com/SurfsUp/Tables.htm)). I'm not sure why this is happening.  The table passes as valid HTML.  Jaws just isn't switching out Virtual PC Cursor Mode and into Table Layer Mode.  Need help investigating if any custom scripted widgets on the page in which the table is displayed are causing Jaws not to make the switch.

packages/terra-clinical-result/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
   * Added screen-reader support for clinical-result icons. (Requires Jest test updates on consuming applications)
   * Added screen-reader support for alternative text for when no results are returned.
   * Added screen-reader support for clinical-result FlowsheetResultCell EnteredInError.
+  * Changed FlowsheetResultCell, ResultNameHeaderCell and ResultTimeHeaderCell in clinical-result to use proper semantic html
```
sdadn, line 9:
```suggestion
  * Changed FlowsheetResultCell, ResultNameHeaderCell and ResultTimeHeaderCell in clinical-result to use proper semantic html.
```

sdadn:
Can you rebase with main so that this PR is up to date?

packages/terra-clinical-result/CHANGELOG.md
RayGunY: Fixed: https://github.com/cerner/terra-clinical/pull/865/files#diff-2f347c01ccae568264be55304c9e1799548817ccb16246b00f7baa02940f2e05R15

sdadn (approved):

RayGunY:
@brooksienoodlesoup in regards to this note you left in your comment

> NOTE - In order to test table headers or table data cells, I can only really test them with a screen reader when they are combined into a full fledged table that has both, like you set up with the Semantic Table with Flowsheet Result Cell.

I went ahead and adjusted the site tests for ResultNameHeaderCell and ResultTimeHeaderCell to include flowsheet result cells as well. I also went into the FlowsheetResultCell site test and replaced the generic HeaderCell component I was using with the ResultNameHeaderCell component. So now all three of those tests should have a combination of using a flowsheet result cell and a time or name header cell. Note that in these table examples not all of the cells are using flowsheetResultCell - wherever you see a blood pressure result that is using the flowsheet result cell, otherwise it's just using '<Cell>'. 

As for your issue with the table not being read automatically with JAWs I believe that's an issue with the terra-core table component since that's what I'm using for these examples. I don't think it should hold up this PR being merged.

RayGunY:
@sdadn @sycombs went ahead and rebased this again since some stuff got merged

sycombs:
This looks good to me, just looks like it needs UX review before it can be merged.

brooksienoodlesoup (approved):

RayGunY:
> This looks good to me, just looks like it needs UX review before it can be merged.

Ah! Ok didn't realize that was still pending lol
Looks like Brooks just approved from UX

PR #868, opened 2023-06-09
[terra-clinical-result] Restructure the Blood Pressure with additional displays as a group

vinaybhargavar:
### Summary

When we have additional displays (concept, DateTime) as part of blood pressure results, we will need to restructure them so that the screen reader understands them as a group.

**What was changed:**
* A new prop was added to allow the consumers to make the blood pressure to be grouped. When the prop is set to true, the blood pressure results will be understood as a group. The group name/label is set to "Blood Pressure {dateTime}". 

**Why it was changed:**
When there are additional displays along with the blood pressure results, the accessibility tooling can recognize the cell as a group context.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

When the prop is false:

https://github.com/cerner/terra-clinical/assets/55503680/6dbf8784-2222-4199-b432-66bc5b68f36f

When the prop is true:

https://github.com/cerner/terra-clinical/assets/55503680/ffd2efde-6d7b-4ae0-bd17-6cd7876c7245

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9096

---

@cerner/terra

### DIFF
--- a/packages/terra-clinical-result/package.json
+++ b/packages/terra-clinical-result/package.json
@@ -36,7 +36,8 @@
     "terra-icon": "^3.53.0",
     "terra-mixins": "^1.0.0",
     "terra-theme-context": "^1.0.0",
-    "terra-visually-hidden-text": "^2.36.0"
+    "terra-visually-hidden-text": "^2.36.0",
+    "uuid": "8.2.0"
   },
   "scripts": {
     "compile": "babel --root-mode upward src --out-dir lib --copy-files",
--- a/packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
+++ b/packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
@@ -6,7 +6,9 @@ import ThemeContext from 'terra-theme-context';
 import IconModified from 'terra-icon/lib/icon/IconModified';
 import IconComment from 'terra-icon/lib/icon/IconComment';
 import IconUnverified from 'terra-icon/lib/icon/IconDiamond';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
 import { injectIntl } from 'react-intl';
+import { v4 as uuidv4 } from 'uuid';
 import observationPropShape from './proptypes/observationPropTypes';
 import ResultError from './common/other/_ResultError';
 import NoData from './common/other/_KnownNoData';
@@ -46,6 +48,10 @@ const propTypes = {
    */
   hasResultNoData: PropTypes.bool,
   /**
+   * Whether or not the result should be a group.
+   */
+  isBloodPressureGrouped: PropTypes.bool,
+  /**
    * @private
    * Used by Flowsheet Result Cell to hide icons because it displays them in different positions.
    */
@@ -61,6 +67,7 @@ const defaultProps = {
   isTruncated: false,
   hasResultError: false,
   hasResultNoData: false,
+  isBloodPressureGrouped: false,
   hideAccessoryDisplays: false,
 };
 
@@ -84,13 +91,14 @@ const createConceptDisplays = (compareConceptDisplays) => {
   return null;
 };
 
+const idForDatetimeDisplays = `${uuidv4()}-datetimeDisplay`;
 const createDatetimeDisplays = (compareDatetimeDisplays) => {
   if (compareDatetimeDisplays.systolic && compareDatetimeDisplays.diastolic) {
     if (compareDatetimeDisplays.systolic === compareDatetimeDisplays.diastolic) {
-      return <div className={cx('datetime-display')}>{compareDatetimeDisplays.originalSystolic}</div>;
+      return <div className={cx('datetime-display')} id={idForDatetimeDisplays}>{compareDatetimeDisplays.originalSystolic}</div>;
     }
     return (
-      <div className={cx('datetime-display')}>
+      <div className={cx('datetime-display')} id={idForDatetimeDisplays}>
         {compareDatetimeDisplays.originalSystolic}
         {' / '}
         {compareDatetimeDisplays.originalDiastolic}
@@ -99,7 +107,7 @@ const createDatetimeDisplays = (compareDatetimeDisplays) => {
   }
   if (compareDatetimeDisplays.systolic || compareDatetimeDisplays.diastolic) {
     const conceptDisplayValue1 = compareDatetimeDisplays.originalSystolic || compareDatetimeDisplays.originalDiastolic;
-    return <div className={cx('datetime-display')}>{conceptDisplayValue1}</div>;
+    return <div className={cx('datetime-display')} id={idForDatetimeDisplays}>{conceptDisplayValue1}</div>;
   }
   return null;
 };
@@ -113,6 +121,7 @@ const ClinicalResultBloodPressure = (props) => {
     isTruncated,
     hasResultError,
     hasResultNoData,
+    isBloodPressureGrouped,
     hideAccessoryDisplays,
     intl,
     ...customProps
@@ -175,10 +184,16 @@ const ClinicalResultBloodPressure = (props) => {
     { 'status-in-error': systolicResult.statusInError || diastolicResult.statusInError },
   ]);
 
+  const idForHiddenText = ((isBloodPressureGrouped) ? `${uuidv4()}-hiddenText` : undefined);
+
   const clinicalResultBloodPressureDisplay = (
-    <React.Fragment>
+    <span
+      role={isBloodPressureGrouped ? 'group' : undefined}
+      aria-labelledby={isBloodPressureGrouped ? `${idForHiddenText} ${idForDatetimeDisplays}` : undefined}
+    >
       <div className={decoratedResultClassnames}>
         <div className={cx('result-display')}>
+          {isBloodPressureGrouped && <VisuallyHiddenText id={idForHiddenText} aria-hidden="true" text={intl.formatMessage({ id: 'Terra.clinicalResult.bloodPressure' })} />}
           {decoratedResultDisplay}
           {isTruncated ? null : !hideAccessoryDisplays && iconGroupDisplayElement}
         </div>
@@ -186,7 +201,7 @@ const ClinicalResultBloodPressure = (props) => {
       </div>
       {!hideAccessoryDisplays && conceptDisplayElement}
       {!hideAccessoryDisplays && datetimeDisplayElement}
-    </React.Fragment>
+    </span>
   );
 
   const clinicalResultClassnames = classNames(
--- a/packages/terra-clinical-result/tests/jest/ClinicalResultBloodPressure.test.jsx
+++ b/packages/terra-clinical-result/tests/jest/ClinicalResultBloodPressure.test.jsx
@@ -6,8 +6,11 @@ import {
   DefaultBloodPressureResult,
   DefaultSystolicResult,
   DefaultDiastolicResult,
+  ExtraDisplaysBloodPressureResult,
   NoDataResult,
 } from '../../src/terra-dev-site/test/clinical-result/TestResults';
+
+jest.mock('uuid', () => ({ v4: () => '00000000-0000-0000-0000-000000000000' }));
 
 describe('ClinicalResultBloodPressure', () => {
   it('should render a ResultError if hasResultError is true', () => {
@@ -18,6 +21,22 @@ describe('ClinicalResultBloodPressure', () => {
   it('should render a NoData if hasResultNoData is true', () => {
     const result = shallowWithIntl(<ClinicalResultBloodPressure hasResultNoData />).dive();
     expect(result).toMatchSnapshot();
+  });
+
+  it('should render ClinicalResultBloodPressure if isBloodPressureGrouped is true', () => {
+    const result = shallowWithIntl(<ClinicalResultBloodPressure {...DefaultBloodPressureResult} isBloodPressureGrouped />).dive();
+    expect(result.find('VisuallyHiddenText').at(0).prop('text')).toEqual('Terra.clinicalResult.bloodPressure');
+    expect(result).toMatchSnapshot();
+  });
+
+  it('should set the hidden text ID', () => {
+    const result = shallowWithIntl(<ClinicalResultBloodPressure {...DefaultBloodPressureResult} isBloodPressureGrouped />).dive();
+    expect(result.find('VisuallyHiddenText').prop('id')).toEqual('00000000-0000-0000-0000-000000000000-hiddenText');
+  });
+
+  it('should set the dateTime display ID', () => {
+    const result = shallowWithIntl(<ClinicalResultBloodPressure {...ExtraDisplaysBloodPressureResult} />).dive();
+    expect(result.find('.datetime-display').prop('id')).toEqual('00000000-0000-0000-0000-000000000000-datetimeDisplay');
   });
 
   it('should render a default ClinicalResultBloodPressure', () => {
--- a/packages/terra-clinical-result/translations/en-AU.json
+++ b/packages/terra-clinical-result/translations/en-AU.json
@@ -17,5 +17,6 @@
   "Terra.clinicalResult.interpretationPositive": "Positive",
   "Terra.clinicalResult.interpretationAbnormal": "Abnormal",
   "Terra.clinicalResult.interpretationHigh": "High",
-  "Terra.clinicalResult.interpretationLow": "Low"
+  "Terra.clinicalResult.interpretationLow": "Low",
+  "Terra.clinicalResult.bloodPressure": "Blood Pressure"
 }
--- a/packages/terra-clinical-result/translations/fi-FI.json
+++ b/packages/terra-clinical-result/translations/fi-FI.json
@@ -17,5 +17,6 @@
   "Terra.clinicalResult.interpretationPositive": "Positive",
   "Terra.clinicalResult.interpretationAbnormal": "Abnormal",
   "Terra.clinicalResult.interpretationHigh": "High",
-  "Terra.clinicalResult.interpretationLow": "Low"
+  "Terra.clinicalResult.interpretationLow": "Low",
+  "Terra.clinicalResult.bloodPressure": "Blood Pressure"
 }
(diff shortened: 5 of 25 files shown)


vinaybhargavar:
@RayGunY, @zxeleven  - I will update the PR description with the relevant details but the code is ready for review.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -177,15 +189,21 @@ const ClinicalResultBloodPressure = (props) => {
...
-        <div className={cx('result-display')}>
-          {decoratedResultDisplay}
-          {isTruncated ? null : !hideAccessoryDisplays && iconGroupDisplayElement}
+      <span
+        role={IsBloodPressureAGroup ? 'group' : undefined}
+        aria-labelledby={IsBloodPressureAGroup ? `id-hidden-text ${idForDatetimeDisplays}` : undefined}
+      >
+        <div className={decoratedResultClassnames}>
+          <div className={cx('result-display')}>
+            {IsBloodPressureAGroup && <VisuallyHiddenText id="id-hidden-text" aria-hidden="true" text="Blood Pressure" />}
```
vinaybhargavar, line 198: I need to update the hardcoded string to be from the translation file.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -45,6 +48,11 @@ const propTypes = {
    * Whether or not the result has a value for a specific datetime.
    */
   hasResultNoData: PropTypes.bool,
+  /**
+   * /**
+   * Whether or not the result should be a group.
+   */
+  IsBloodPressureAGroup: PropTypes.bool,
```
RayGunY, line 55: Nit: lowercase the 'i' and maybe we could rename this to isBloodPressureGrouped - what do you think?

vinaybhargavar: isBloodPressureGrouped sounds better. I can rename the prop.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -177,15 +189,21 @@ const ClinicalResultBloodPressure = (props) => {
 
   const clinicalResultBloodPressureDisplay = (
     <React.Fragment>
-      <div className={decoratedResultClassnames}>
-        <div className={cx('result-display')}>
-          {decoratedResultDisplay}
-          {isTruncated ? null : !hideAccessoryDisplays && iconGroupDisplayElement}
+      <span
+        role={IsBloodPressureAGroup ? 'group' : undefined}
```
RayGunY, line 193: Do you happen to know the difference between setting this to undefined vs null? I was trying to poke around for more info about how this is handled in relationship to the aria roles but didn't find much - I've seen it done both ways. In general my understanding is that null is used to define something that is programmatically empty and undefined is meant to say that the reference is not existing. I'm pretty sure it shouldn't matter functionality wise but do we care what the role should be associated with?

 i.e. when IsBloodPressureAGroup is false do we want this role to be set to null and therefore implied that it is intentionally a defined reference set to 'nothing' or do we want to set it to undefined which implies that there is no existing property to reference?

I'm definitely overcomplicating this but just wanted to know if you had any thoughts?

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -45,6 +48,11 @@ const propTypes = {
    * Whether or not the result has a value for a specific datetime.
    */
   hasResultNoData: PropTypes.bool,
+  /**
+   * /**
```
zxeleven, line 52: nit: remove this line

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -84,13 +93,15 @@ const createConceptDisplays = (compareConceptDisplays) => {
   return null;
 };
 
+let idForDatetimeDisplays;
 const createDatetimeDisplays = (compareDatetimeDisplays) => {
+  idForDatetimeDisplays = `${uuidv4()}-dateTimeDisplay`;
```
zxeleven, line 98: small nit: `dateTimeDisplay` -> `datetimeDisplay`

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
RayGunY: For this line if aria-hidden is true for the VisuallyHiddenText then we won't see it on the front end and the screenreader will skip it - so it wouldn't be seen or heard. Is that the expected behavior for this?

vinaybhargavar: Yup, we need this to be read only when the prop is set to true. If the prop is false, then just the result needs to be read out. 
@brooksienoodlesoup - Can you confirm how the behavior of this should be?

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
vinaybhargavar: Updated here - https://github.com/cerner/terra-clinical/pull/868/commits/ba6cb16362db09a115cd6bb88d36f7853bfabc10#diff-42afe5315e2b63b43f755f1ec567ef32b1ad903109d718db4e7526f825680f6aR54

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
vinaybhargavar: Changed here - https://github.com/cerner/terra-clinical/pull/868/commits/ba6cb16362db09a115cd6bb88d36f7853bfabc10#diff-42afe5315e2b63b43f755f1ec567ef32b1ad903109d718db4e7526f825680f6aR97

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
vinaybhargavar: https://github.com/cerner/terra-clinical/pull/868/commits/ba6cb16362db09a115cd6bb88d36f7853bfabc10#diff-42afe5315e2b63b43f755f1ec567ef32b1ad903109d718db4e7526f825680f6aR51

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
vinaybhargavar: Updated to use translation string - https://github.com/cerner/terra-clinical/pull/868/commits/ba6cb16362db09a115cd6bb88d36f7853bfabc10#diff-42afe5315e2b63b43f755f1ec567ef32b1ad903109d718db4e7526f825680f6aR197. 

Left a comment on Reagan's [Jira](https://jira.cerner.com/browse/TRANS-14531?focusedCommentId=4752432&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4752432) asking if they need a new Jira or if they can translate to non-english ones as part of the same Jira.

vinaybhargavar:
Dev site - https://engineering.cerner.com/terra-clinical/pull/868

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
vinaybhargavar: I think both should work the same but I haven't tested thoroughly with role= null. I was checking the code base and on the [data grid component](https://github.com/cerner/terra-clinical/blob/23aec842dc94c704243ddfcaa6a0883946ad9cde/packages/terra-clinical-data-grid/src/DataGrid.jsx#L1248) I saw that they had used undefined for aria-labelledby property. Hence I went ahead with undefined here as well.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
vinaybhargavar: This will be done as part of Reagan's story - https://jira.cerner.com/browse/TRANS-14531?focusedCommentId=4752846&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4752846

RayGunY: I think I'm still confused, here even when isBloodPressureGrouped true the text won't be seen or read out

vinaybhargavar: The text 'Blood Pressure' is never going to be seen irrespective of whether isBloodPressureGrouped is true or false. If isBloodPressureGrouped is true, then the screen reader should read it.

RayGunY: Yup that first part makes sense about it not being seen, but the screenreader won't read it out at all if aria-hidden is set to true. Right? That was my understanding: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

brooksienoodlesoup (approved):
Tested in:
- Chrome 108.0.5359.94 (Official Build) (arm64) + VO (Ventura 13.2.1)
- Edge 114.0.1823.43 (Official build) (64-bit) + JAWS 2023.2303.144

Reads out as expected by both JAWS and VO.

zxeleven (approved):

RayGunY (approved):

packages/terra-clinical-result/package.json
```diff
@@ -36,7 +36,8 @@
     "terra-icon": "^3.53.0",
     "terra-mixins": "^1.0.0",
     "terra-theme-context": "^1.0.0",
-    "terra-visually-hidden-text": "^2.36.0"
+    "terra-visually-hidden-text": "^2.36.0",
+    "uuid": "^3.4.0"
```
sdadn, line 40: Can we update and lock this to `8.2.0` since that is what's being used across Terra?
```suggestion
    "uuid": "8.2.0"
```

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -6,7 +6,10 @@ import ThemeContext from 'terra-theme-context';
 import IconModified from 'terra-icon/lib/icon/IconModified';
 import IconComment from 'terra-icon/lib/icon/IconComment';
 import IconUnverified from 'terra-icon/lib/icon/IconDiamond';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
 import { injectIntl } from 'react-intl';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import uuidv4 from 'uuid/v4';
```
sdadn, line 12: This needs to be updated to use the import from uuid@8 once that is pulled in.

packages/terra-clinical-result/translations/en-AU.json
```diff
@@ -17,5 +17,6 @@
   "Terra.clinicalResult.interpretationPositive": "Positive",
   "Terra.clinicalResult.interpretationAbnormal": "Abnormal",
   "Terra.clinicalResult.interpretationHigh": "High",
-  "Terra.clinicalResult.interpretationLow": "Low"
+  "Terra.clinicalResult.interpretationLow": "Low",
+  "Terra.clinicalResult.bloodPressure": "Blood Pressure"
```
sdadn, line 21: Is there a translations Jira to update these translations in the future?

packages/terra-clinical-result/translations/fi-FI.json
```diff
@@ -17,5 +17,6 @@
   "Terra.clinicalResult.interpretationPositive": "Positive",
   "Terra.clinicalResult.interpretationAbnormal": "Abnormal",
   "Terra.clinicalResult.interpretationHigh": "High",
-  "Terra.clinicalResult.interpretationLow": "Low"
+  "Terra.clinicalResult.interpretationLow": "Low",
+  "Terra.clinicalResult.bloodPressure": "Blood Pressure"
```
sdadn, line 21: That's weird that the Finnish translations are in English, though fixing it is beyond the scope of this PR.

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -175,17 +186,25 @@ const ClinicalResultBloodPressure = (props) => {
     { 'status-in-error': systolicResult.statusInError || diastolicResult.statusInError },
   ]);
 
+  const idForHiddenText = ((isBloodPressureGrouped) ? `${uuidv4()}-hiddenText` : undefined);
+
   const clinicalResultBloodPressureDisplay = (
     <React.Fragment>
```
sdadn, line 192: Since we're only have a single element, we no longer need to use a `React.Fragment`.

```suggestion
```

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
```diff
@@ -84,13 +92,15 @@ const createConceptDisplays = (compareConceptDisplays) => {
   return null;
 };
 
+let idForDatetimeDisplays;
 const createDatetimeDisplays = (compareDatetimeDisplays) => {
+  idForDatetimeDisplays = `${uuidv4()}-datetimeDisplay`;
```
sdadn, line 97: We can assign `idForDatetimeDisplays` upon creation as it is not being mutated.
```suggestion
const idForDatetimeDisplays = `${uuidv4()}-datetimeDisplay`;
const createDatetimeDisplays = (compareDatetimeDisplays) => {
```

sdadn (changes requested):

packages/terra-clinical-result/tests/jest/ClinicalResultBloodPressure.test.jsx
```diff
@@ -9,6 +9,8 @@ import {
   NoDataResult,
 } from '../../src/terra-dev-site/test/clinical-result/TestResults';
 
+jest.mock('uuid/v4', () => () => '55b406c0-057c-4e82-94fb-85a2fb87c29c');
```
sdadn, line 12: There should be a test to see if uuid is being set. We can also use `00000000-0000-0000-0000-000000000000` as the mock value for better maintainability.

sdadn (changes requested):

packages/terra-clinical-result/translations/en-AU.json
vinaybhargavar: Yup, the translations would be done as part of TRANS-14531. See this [comment](https://jira.cerner.com/browse/TRANS-14531?focusedCommentId=4752846&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4752846).

packages/terra-clinical-result/package.json
vinaybhargavar: Updated here - https://github.com/cerner/terra-clinical/pull/868/commits/f932e12eaad49881c42630811d1619d1000807ed#diff-9084f5bebd8f555fe623f1b8c564a60cd145af644151dd4f498daa620d7fd87fR40. But I see warnings when I run lint. Is that fine?
Screen Shot 2023-06-16 at 7 45 23 PM

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
vinaybhargavar: Updated here - https://github.com/cerner/terra-clinical/pull/868/commits/f932e12eaad49881c42630811d1619d1000807ed#diff-42afe5315e2b63b43f755f1ec567ef32b1ad903109d718db4e7526f825680f6aR11

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
vinaybhargavar: Removed here - https://github.com/cerner/terra-clinical/pull/868/commits/f932e12eaad49881c42630811d1619d1000807ed#diff-42afe5315e2b63b43f755f1ec567ef32b1ad903109d718db4e7526f825680f6aR190

packages/terra-clinical-result/src/ClinicalResultBloodPressure.jsx
vinaybhargavar: Updated here - https://github.com/cerner/terra-clinical/pull/868/commits/f932e12eaad49881c42630811d1619d1000807ed#diff-42afe5315e2b63b43f755f1ec567ef32b1ad903109d718db4e7526f825680f6aR94

packages/terra-clinical-result/tests/jest/ClinicalResultBloodPressure.test.jsx
vinaybhargavar: Added tests here - https://github.com/cerner/terra-clinical/pull/868/commits/f932e12eaad49881c42630811d1619d1000807ed#diff-f78e39f6ea1427cf52ee7390619f232d70376b20b4e2739af9e9f78b662cc681R32 and updated to use `00000000-0000-0000-0000-000000000000` here - https://github.com/cerner/terra-clinical/pull/868/commits/f932e12eaad49881c42630811d1619d1000807ed#diff-f78e39f6ea1427cf52ee7390619f232d70376b20b4e2739af9e9f78b662cc681R13

packages/terra-clinical-result/translations/en-AU.json
vinaybhargavar: Updated the translations for non English languages - https://github.com/cerner/terra-clinical/pull/868/commits/3f28c70e6aa77ceda8d00cb3a50ae80f362ce9b5

packages/terra-clinical-result/translations/fi-FI.json
vinaybhargavar: I didnt find any translations for Finnish, I have asked the same on the Jira - https://jira.cerner.com/browse/TRANS-14531?focusedCommentId=4755676&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4755676

sdadn (approved):

packages/terra-clinical-result/translations/fi-FI.json
sdadn: That's fine! We'll follow up on this on our end; We may no longer be supporting Finnish. Thanks btw!

vinaybhargavar: Thanks @sdadn

PR #870, opened 2023-06-13
[terra-clinical-result] Add alt text for strikethroughs to be understood as entered in error

RayGunY:
### Summary
When we have a result with a strikethrough that means to the visual user that the result has been deleted in some way, in our case that it has been entered in error. We need to convey a message that tells the screenreader user that the result is no longer relevant and has been 'Entered in error'. That message would look like "Result {someResult} Entered In Error". Note: translations for this message are still being figured out.

**What was changed:**

The initial strikethrough result is now aria-hidden and instead of using "< del>" we'll be using a regular strikethrough "< s>". This is so the screenreader doesn't read out 'deleted' along with the result and instead just reads out our custom message. 
Then a visually hidden text element has been added to the conditional wrapper after the strikethrough element for the result. To populate the text in that visually hidden element I opted to directly use the [result](https://github.com/cerner/terra-clinical/blob/main/packages/terra-clinical-result/src/_ClinicalResultDisplay.jsx#L100) being passed in and not the children. The children in this case are what observation will return, which is an [object containing up to two spans](https://github.com/cerner/terra-clinical/blob/main/packages/terra-clinical-result/src/common/observation/_Observation.jsx#L99-L106)(the unit span is optional). We don't need the spans, just their text and the [result.value and result.unit](https://github.com/cerner/terra-clinical/blob/main/packages/terra-clinical-result/src/common/observation/_Observation.jsx#L104C11-L106) is what's populating those anyway. 

To use the result directly, before rendering the hidden text I added a check for if the result.value is valid using the same logic as [here](https://github.com/cerner/terra-clinical/blob/main/packages/terra-clinical-result/src/common/observation/_Observation.jsx#L76). I also added a check for if the unit exists, like [this](https://github.com/cerner/terra-clinical/blob/main/packages/terra-clinical-result/src/common/observation/_Observation.jsx#L106) as well as a check for if hideUnit is set to true (in which case we don't want to read out the unit even if it exists).

If the result value is not valid then we just return null and do not render the visually hidden text at all. This is because in the observation file when the result value isn't valid it returns a [result error](https://github.com/cerner/terra-clinical/blob/main/packages/terra-clinical-result/src/common/observation/_Observation.jsx#L110) so we don't want to read out anything extra if that's the case. 

**Why it was changed:**
When there is a strikethrough it's currently reading out the result and that it's been deleted but that's not consistent among all the browsers/tools. We also want to give greater context to the strikethrough result and the fact that it is not just deleted but entered in error. Lots of discussion has gone on about what the exact messaging should be, but we think we've landed on the current string I mentioned in the summary - more info can be found in the comments of the jira tied to this PR. 

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9067

---

@cerner/terra

Clinical Result:

https://github.com/cerner/terra-clinical/assets/40574651/c5b939bb-66f0-41f9-b250-1d0b3d59612f

Clinical Result Blood Pressure:

https://github.com/cerner/terra-clinical/assets/40574651/74865bdc-3a53-4148-8381-21c7379b473b

### DIFF
--- a/packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
+++ b/packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { injectIntl } from 'react-intl';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
 import observationPropShape from './proptypes/observationPropTypes';
 import Observation from './common/observation/_Observation';
 import ResultError from './common/other/_ResultError';
@@ -27,6 +29,11 @@ const propTypes = {
    * The cleaned unit of the diastolic display.
    */
   diastolicUnit: PropTypes.string,
+  /**
+   * @private
+   * The intl object to be injected for translations.
+   */
+  intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,
 };
 
 const BloodPressureDisplay = ({
@@ -35,7 +42,10 @@ const BloodPressureDisplay = ({
   id,
   type,
   diastolicUnit,
+  intl,
 }) => {
+  const isValidValue = result?.result?.value;
+
   if (!result) {
     return <ResultError key={`Error-${type}-${id}`} />;
   }
@@ -46,7 +56,20 @@ const BloodPressureDisplay = ({
     <ConditionalWrapper
       key={`del-${type}-${result.eventId}`}
       condition={result.statusInError}
-      wrapper={children => <del>{children}</del>}
+      wrapper={children => (
+        <span>
+          <s aria-hidden="true">{children}</s>
+          {isValidValue
+            ? (
+              <VisuallyHiddenText text={intl.formatMessage(
+                { id: 'Terra.clinicalResult.statusInErrorStrikethrough' },
+                { strikethroughResult: `${result.result.value} ${!hideUnit && result.result.unit ? result.result.unit : ''}` },
+              )}
+              />
+            )
+            : null}
+        </span>
+      )}
     >
       <Observation
         key={`Observation-${type}-${result.eventId}`}
@@ -62,4 +85,4 @@ const BloodPressureDisplay = ({
 
 BloodPressureDisplay.propTypes = propTypes;
 
-export default BloodPressureDisplay;
+export default injectIntl(BloodPressureDisplay);
--- a/packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
+++ b/packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
@@ -45,6 +45,7 @@ const BloodPressureDisplay = ({
   intl,
 }) => {
   const isValidValue = result?.result?.value;
+  const isUnitPresent = !hideUnit && result.result.unit;
 
   if (!result) {
     return <ResultError key={`Error-${type}-${id}`} />;
@@ -63,7 +64,7 @@ const BloodPressureDisplay = ({
             ? (
               <VisuallyHiddenText text={intl.formatMessage(
                 { id: 'Terra.clinicalResult.statusInErrorStrikethrough' },
-                { strikethroughResult: `${result.result.value} ${!hideUnit && result.result.unit ? result.result.unit : ''}` },
+                { strikethroughResult: `${result.result.value} ${isUnitPresent ? result.result.unit : ''}` },
               )}
               />
             )
--- a/packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
+++ b/packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
@@ -45,7 +45,7 @@ const BloodPressureDisplay = ({
   intl,
 }) => {
   const isValidValue = result?.result?.value;
-  const isUnitPresent = !hideUnit && result.result.unit;
+  const isUnitPresent = !hideUnit && result?.result?.unit;
 
   if (!result) {
     return <ResultError key={`Error-${type}-${id}`} />;
--- a/packages/terra-clinical-result/src/_ClinicalResultDisplay.jsx
+++ b/packages/terra-clinical-result/src/_ClinicalResultDisplay.jsx
@@ -88,6 +88,7 @@ const ClinicalResultDisplay = ({
   intl,
 }) => {
   const isValidValue = result?.value;
+  const isUnitPresent = !hideUnit && result.unit;
   const isStatusInError = !isEmpty(status) ? checkIsStatusInError(status) : false;
   const decoratedResultClassnames = cx([
     'decorated-result-display',
@@ -108,7 +109,7 @@ const ClinicalResultDisplay = ({
                   ? (
                     <VisuallyHiddenText text={intl.formatMessage(
                       { id: 'Terra.clinicalResult.statusInErrorStrikethrough' },
-                      { strikethroughResult: `${result.value} ${!hideUnit && result.unit ? result.unit : ''}` },
+                      { strikethroughResult: `${result.value} ${isUnitPresent ? result.unit : ''}` },
                     )}
                     />
                   )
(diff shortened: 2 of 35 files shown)


packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
```diff
@@ -46,7 +56,20 @@ const BloodPressureDisplay = ({
...
       condition={result.statusInError}
-      wrapper={children => <del>{children}</del>}
+      wrapper={children => (
+        <span>
+          <s aria-hidden="true">{children}</s>
+          {isValidValue
+            ? (
+              <VisuallyHiddenText text={intl.formatMessage(
+                { id: 'Terra.clinicalResult.statusInErrorStrikethrough' },
+                { strikethroughResult: `${result.result.value} ${result.result.unit ? result.result.unit : ''}` },
```
chenyixuan625, line 66: Just curious, if there is no unit for the result, will the screen reader read out anything else after the value? (Basically want to confirm if the empty string would cause any unexpected behavior or not)

RayGunY: Great question, the space doesn't cause any issues - however I will mention that I am currently adding a check here for if hideUnit is true - we won't want to read out the unit if it's not being show.

RayGunY:
I added some terra site tests for clinical result (result with unit and result without unit) and clinical blood pressure result (blood pressure result without unit and blood pressure result with partial entered in error status)

https://github.com/cerner/terra-clinical/assets/40574651/ba573eb1-0133-45e6-b84a-f275dda031d0

https://github.com/cerner/terra-clinical/assets/40574651/ce095013-17af-49d2-ba14-e9a8c3c221f6

https://github.com/cerner/terra-clinical/assets/40574651/d30dcb81-b4ef-4156-8c6f-5c99bdeb3ba4

https://github.com/cerner/terra-clinical/assets/40574651/8ade2e4b-643d-4133-b9a7-b66f9a32ecb4

RayGunY:
I just pushed a commit that updates the name of a translation mostly unrelated to this story. I noticed when looking in the translation file that the translation for our in error component was still named '.statusInErrorAria' - this is no longer accurate as the string is not being used in aria anymore but in a hidden text component. More details on that can be found in this PR: https://jira2.cerner.com/browse/UXPLATFORM-9065

Talked with Krystal and she asked if I could go ahead and update it alongside these changes. 

Commit found here: [ccf695c](https://github.com/cerner/terra-clinical/pull/870/commits/ccf695c432ec48edcab255f6db6c28eb27671ffb)

sycombs (approved):

sdadn (approved):
Looks good!

packages/terra-clinical-result/src/terra-dev-site/test/clinical-result/clinical-result-blood-pressure/EnteredInErrorClinicalResultBloodPressureNoUnit.test.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { ClinicalResultBloodPressure } from 'terra-clinical-result/lib/index';
+
+const resultData = {
+  systolic: {
+    eventId: '111.1',
+    result: {
+      value: '180',
+      unit: 'mmHg',
```
kolkheang, line 9: Could you add a test where `unit` isn't present in the data?

packages/terra-clinical-result/src/_ClinicalResultDisplay.jsx
```diff
@@ -78,7 +85,9 @@ const ClinicalResultDisplay = ({
   hasComment,
   conceptDisplay,
   datetimeDisplay,
+  intl,
 }) => {
+  const isValidValue = result?.value;
```
kolkheang, line 90: How is the incoming `result` prop data look like? Is it the same as the one above? `result?.result?.value` ?

RayGunY: Between ClinicalResultDisplay and BloodPressureDisplay they are a little different props wise, you can see how in blood pressure display they are passing in result.result in order to then be able to access the value and unit in the observation file. You can see result.result being passed in here:https://github.com/cerner/terra-clinical/blob/ccf695c432ec48edcab255f6db6c28eb27671ffb/packages/terra-clinical-result/src/_BloodPressureDisplay.jsx#L77

For clinical result display they are just passing in result so result.result.value is not necessary in this case: https://github.com/cerner/terra-clinical/blob/ccf695c432ec48edcab255f6db6c28eb27671ffb/packages/terra-clinical-result/src/_ClinicalResultDisplay.jsx#L122

packages/terra-clinical-result/src/terra-dev-site/test/clinical-result/clinical-result-blood-pressure/EnteredInErrorClinicalResultBloodPressureNoUnit.test.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
```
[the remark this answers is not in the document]
RayGunY: Sure thing! So I renamed this file to be EnteredInErrorClinicalResultBloodPressureHiddenUnit and then added another test for when unit isn't present at all and named it EnteredInErrorClinicalResultBloodPressureNoUnit. I also did the same with the Clinical result tests.

packages/terra-clinical-result/src/terra-dev-site/test/clinical-result/clinical-result-blood-pressure/EnteredInErrorClinicalResultBloodPressureNoUnit.test.jsx
RayGunY: Changed tests around here, added two tests for unit not being present - one for blood pressure result and one for clinical result: [632b661](https://github.com/cerner/terra-clinical/pull/870/commits/632b661f37653bb9f89d4c4ddb50cfb172613676)

packages/terra-clinical-result/src/_BloodPressureDisplay.jsx
```diff
@@ -46,7 +56,20 @@ const BloodPressureDisplay = ({
...
       condition={result.statusInError}
-      wrapper={children => <del>{children}</del>}
+      wrapper={children => (
+        <span>
+          <s aria-hidden="true">{children}</s>
+          {isValidValue
+            ? (
+              <VisuallyHiddenText text={intl.formatMessage(
+                { id: 'Terra.clinicalResult.statusInErrorStrikethrough' },
+                { strikethroughResult: `${result.result.value} ${!hideUnit && result.result.unit ? result.result.unit : ''}` },
```
chenyixuan625, line 66: Nit: Since the hideUnit check was added, would you mind pulling the whole expression out and making it a variable?

RayGunY: Good idea, changed here: [6c8a6ad](https://github.com/cerner/terra-clinical/pull/870/commits/6c8a6add181de629fe965e6dbc78fee621666359)

brooksienoodlesoup (approved):
Tested with:

- Chrome 108.0.5359.94 (Official Build) (arm64) + VO (Ventura 13.2.1)
- Edge 114.0.1823.43 (Official build) (64-bit) + JAWS 2023.2303.144

Screen reader output was as expected.  What is presented visually as strikethrough text is announced by VO and JAWS as regular text, then announced "in error." Nice work!

*NOTE:*  We have to make very, very sure that there are some rock solid checks here that make sure what is being presented to screen reader users as visually hidden text is identical to the data that is passed as strikethrough text visually.  When you use aria-hidden="true", it makes it exceptionally difficult to catch instances where what's being announced to screen reader users differs from what's presented onscreen.

sdadn (approved):

PR #872, opened 2023-06-15
[terra-clinical-result] Update ResultTimeHeaderCell to use <time> html tags

zxeleven:
### Summary

Updated the ResultTimeHeaderCell component to use `<time>` html tags for the date and time.

**What was changed:**
The ResultTimeHeaderCell component is now using `<time>` html tags for the date and time rather than divs. I also added a visually hidden space between the date and time to resolve an issue with JAWS combining the two and reading them incorrectly.

**Why it was changed:**
So the date and time are read to screen reader users in an understandable way.

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] No tests are needed

#### Visual Testing

https://github.com/cerner/terra-clinical/assets/37230460/c6919048-2add-4467-95ba-f1e237f99889

https://github.com/cerner/terra-clinical/assets/37230460/532209a2-267f-4fb8-8fb0-1c837d759fac

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details
Voiceover and JAWS do not officially support the `<time>` html tag yet, so neither of them read both the date and time %100 correctly.

**This PR resolves:**

UXPLATFORM-9077 

---

@cerner/terra

### DIFF
--- a/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
+++ b/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import classNamesBind from 'classnames/bind';
 import ThemeContext from 'terra-theme-context';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
 import styles from './ResultTimeHeaderCell.module.scss';
 
 const cx = classNamesBind.bind(styles);
@@ -62,7 +63,7 @@ const ResultTimeHeaderCell = (props) => {
       className={timeHeaderCellClassnames}
     >
       <time className={dateClassnames}>{date}</time>
-      <p className={cx('hidden-space')}>{'\u00A0'}</p>
+      <VisuallyHiddenText className={cx('hidden-space')} text={'\u00A0'} />
       <time className={cx('time')}>{time}</time>
     </th>
   );
--- a/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
+++ b/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
@@ -62,6 +62,7 @@ const ResultTimeHeaderCell = (props) => {
       className={timeHeaderCellClassnames}
     >
       <time className={dateClassnames}>{date}</time>
+      <p className={cx('hidden-space')}>{'\u00A0'}</p>
       <time className={cx('time')}>{time}</time>
     </th>
   );
--- a/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
+++ b/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
@@ -63,7 +63,7 @@ const ResultTimeHeaderCell = (props) => {
       className={timeHeaderCellClassnames}
     >
       <time className={dateClassnames}>{date}</time>
-      <VisuallyHiddenText className={cx('hidden-space')} text={'\u00A0'} />
+      <VisuallyHiddenText text={'\u00A0'} />
       <time className={cx('time')}>{time}</time>
     </th>
   );
--- a/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
+++ b/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
@@ -9,7 +9,7 @@ const cx = classNamesBind.bind(styles);
 
 const propTypes = {
   /**
-   * Content to be displayed on the first line, typically abbrivated date, e.g. `Dec 12, 2010`
+   * Content to be displayed on the first line, typically abbreviated date, e.g. `Dec 12, 2010`
    */
   date: PropTypes.string.isRequired,
   /**
@@ -61,8 +61,8 @@ const ResultTimeHeaderCell = (props) => {
       {...customProps}
       className={timeHeaderCellClassnames}
     >
-      <div className={dateClassnames}>{date}</div>
-      <div className={cx('time')}>{time}</div>
+      <time className={dateClassnames}>{date}</time>
+      <time className={cx('time')}>{time}</time>
     </th>
   );
 };
--- a/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.module.scss
+++ b/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.module.scss
@@ -39,8 +39,12 @@
       min-width: 10px; // Needed for IE10
     }
 
+    .hidden-space {
+      display: none;
+    }
+
     /* stylelint-disable-next-line selector-max-compound-selectors */
-    .date + .time {
+    .hidden-space + .time {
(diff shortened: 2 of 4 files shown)


packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
```diff
@@ -62,6 +62,7 @@ const ResultTimeHeaderCell = (props) => {
       className={timeHeaderCellClassnames}
     >
       <time className={dateClassnames}>{date}</time>
+      <p className={cx('hidden-space')}>{'\u00A0'}</p>
```
kolkheang, line 65: Does this need to be a <p> tag?

zxeleven: No I think it could be a different tag but wasn't sure what would make the most sense for this

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
```diff
@@ -61,8 +62,9 @@ const ResultTimeHeaderCell = (props) => {
       {...customProps}
       className={timeHeaderCellClassnames}
     >
-      <div className={dateClassnames}>{date}</div>
-      <div className={cx('time')}>{time}</div>
+      <time className={dateClassnames}>{date}</time>
+      <VisuallyHiddenText className={cx('hidden-space')} text={'\u00A0'} />
```
kolkheang, line 66: 💯 
Thanks @zxeleven !

zxeleven: No problem!

kolkheang:
https://engineering.cerner.com/terra-clinical/pull/872/

Click on Components -> Clinical Results -> Result Time Header Cell

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
kolkheang, line 66: Is the className attribute needed for this component if it's hidden?

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.module.scss
```diff
@@ -40,7 +40,7 @@
     }
 
     /* stylelint-disable-next-line selector-max-compound-selectors */
-    .date + .time {
```
kolkheang, line 43: Also, it looks like `.date` deletion will affect the date display?

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
zxeleven: I was using the className [here](https://github.com/cerner/terra-clinical/pull/872/files#diff-e41002d62a16236c97fa2e987e65bbefee808c6bf73b9dcffec4abdf25ce3fc5R43-R44) to maintain this selector that is setting the `margin-top` value on all elements with the `.time` class but I think we can just move this style into that class.

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.module.scss
zxeleven: The selector here is only using the `.date` class to find the elements with the `.time` class to apply this styling to but I think the `margin-top` style here can just be moved into the `.time` class. See https://github.com/cerner/terra-clinical/pull/872#discussion_r1231553665.

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
zxeleven: Removed here: https://github.com/cerner/terra-clinical/pull/872/commits/06e0278063bf39c26cf889b6aaad6714db468455

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.module.scss
zxeleven: Removed this selector here: https://github.com/cerner/terra-clinical/pull/872/commits/06e0278063bf39c26cf889b6aaad6714db468455

brooksienoodlesoup (commented):
Tested in:
Chrome 108.0.5359.94 (Official Build) (arm64) + VO (Ventura 13.2.1)
Edge 114.0.1823.43 (Official build) (64-bit) + JAWS 2023.2303.144
I got mixed results with both VO and JAWS. Neither provided a clear user experience.

With VO: Dates were read correctly (December read in full), but times were read as numbers.
With JAWS: Dates were read correctly.  But times were read in three different formats: military time, regular time (three o'clock) and numbers (twenty three fifty nine).

These are coded correctly, as far as I can determine.  The varying screen reader output makes this a confusing experience, in my opinion.  Not sure where to go with this, since I wasn't a part of the effort to add time elements from the start.

kolkheang:
Unsure what happened but now VO is announcing the date, and time values are text element where previously I didn't hear those things announced

kolkheang:
> Unsure what happened but now VO is announcing the date, and time values as text element where previously I didn't hear those things announced

Ah, perhaps this is the expected behavior if navigating element by element vs read all text using CTRL+OPT+A

chenyixuan625:
Could we test the behavior of VO and JAWS with the current implementation? I feel like the VO and JAWS read the month abbreviation differently with their default settings.

zxeleven:
> Could we test the behavior of VO and JAWS with the current implementation? I feel like the VO and JAWS read the month abbreviation differently with their default settings.

The recordings in the description are testing using the current implementation but as Brooks pointed out [here](https://github.com/cerner/terra-clinical/pull/872#pullrequestreview-1482447120), the screen reader output is very inconsistent since I think neither Voiceover or JAWS directly supports the `<time>` tag.

Since I also wasn't part of the initial discussion to use these tags, we may need some input from @scottwilmarth to determine if this behavior is expected (i.e. we want to update to the semantic elements now for when they are more widely supported in the future) or what the best path forward should be.

sdadn (approved):

sycombs (approved):

chenyixuan625 (approved):
Code changes look good to me. +1 if we are okay with the screen reader differences.

brooksienoodlesoup:
I agree with @zxeleven  on getting feedback from @scottwilmarth on using the time element, which is isn't supported by Jaws and which leads to inconsistent announcement of time content.

sycombs (approved):

scottwilmarth:
Agreed, neither results are exactly what we would like. The JAWs behavior of reading a few different formats is odd for sure, but all of the headers are coded exactly the same, making me think it is JAWs being inconsistent. Additionally, we are injecting more context by using the semantic TIME element over a DIV. Hopefully, within time user agents will better support the TIME element for a better user experience. Giving my plus one for now, knowing there is likely still work ahead.

RayGunY (approved):
Re-approving since Yixuan is out

kolkheang (approved):

kolkheang:
@brooksienoodlesoup  - the 'Accessibility Review Required' label is still set

brooksienoodlesoup (approved):
<time> element coded to specification.

PR #876, opened 2023-06-26
[terra-clinical-result] Remove <time> tags from ResultTimeHeaderCell

RayGunY:
### Summary

When merging changes for the most recent release two of our PRs didn't play together properly and it was missed. We are seeing some issues with spacing in the ResultTimeHeaderCell component. When in a semantic table the time and date are squished together instead of being on separate newlines. This only effects the semantic examples. 

Issue with semantic table:
Screenshot 2023-06-26 at 11 13 52 AM

Looks fine in non-semantic example:
Screenshot 2023-06-26 at 11 14 39 AM

Update: This lead us to seeing several different issues with how the time tag is working. After a long discussion we have decided to remove it entirely.

**What was changed:**
Change the time tags back to divs, also remove the hidden text between the two.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra

Validation:

Non-semantic 

https://github.com/cerner/terra-clinical/assets/40574651/1d5c349f-3e78-4494-830b-356ae9bc5dd4

Semantic table

https://github.com/cerner/terra-clinical/assets/40574651/68dfde7b-5e04-4c9d-ada3-06d1c6ecad5d

### DIFF
--- a/packages/terra-clinical-result/CHANGELOG.md
+++ b/packages/terra-clinical-result/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Changed ResultTimeHeaderCell to remove the `<time>` html tag for the date and time. This was causing issues and technically does not have full screen reader support.
+
 ## 1.18.0 - (June 22, 2023)
 
 * Added
--- a/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
+++ b/packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import classNamesBind from 'classnames/bind';
 import ThemeContext from 'terra-theme-context';
-import VisuallyHiddenText from 'terra-visually-hidden-text';
 import styles from './ResultTimeHeaderCell.module.scss';
 
 const cx = classNamesBind.bind(styles);
@@ -62,9 +61,8 @@ const ResultTimeHeaderCell = (props) => {
       {...customProps}
       className={timeHeaderCellClassnames}
     >
-      <time className={dateClassnames}>{date}</time>
-      <VisuallyHiddenText text={'\u00A0'} />
-      <time className={cx('time')}>{time}</time>
+      <div className={dateClassnames}>{date}</div>
+      <div className={cx('time')}>{time}</div>
     </th>
   );
 };
--- a/packages/terra-clinical-result/tests/jest/result-time-header-cell/__snapshots__/ResultTimeHeaderCell.test.jsx.snap
+++ b/packages/terra-clinical-result/tests/jest/result-time-header-cell/__snapshots__/ResultTimeHeaderCell.test.jsx.snap
@@ -4,19 +4,16 @@ exports[`ResultTimeHeaderCell correctly applies the theme context className 1`] 
 <th
   className="clinical-result-time-header-cell padding-compact orion-fusion-theme"
 >
-  <time
+  <div
     className="date"
   >
     December 31st 1999
-  </time>
-  <VisuallyHiddenText
-    text=" "
-  />
-  <time
+  </div>
+  <div
     className="time"
   >
     10:00 PM
-  </time>
+  </div>
 </th>
 `;
 
@@ -24,19 +21,16 @@ exports[`ResultTimeHeaderCell paddingStyle - should render with compact 1`] = `
 <th
   className="clinical-result-time-header-cell padding-compact"
 >
-  <time
+  <div
     className="date"
   >
     December 31st 1999
-  </time>
-  <VisuallyHiddenText
-    text=" "
-  />
-  <time
+  </div>
+  <div
     className="time"
   >
     10:00 PM
-  </time>
+  </div>
 </th>
 `;
 
@@ -44,19 +38,16 @@ exports[`ResultTimeHeaderCell paddingStyle - should render with none 1`] = `
 <th
   className="clinical-result-time-header-cell"
 >
-  <time
+  <div
     className="date"
   >
     December 31st 1999
-  </time>
-  <VisuallyHiddenText
-    text=" "
-  />
-  <time
+  </div>
+  <div
     className="time"
   >
     10:00 PM
-  </time>
+  </div>
 </th>
 `;
 
@@ -64,19 +55,16 @@ exports[`ResultTimeHeaderCell paddingStyle - should render with standard 1`] = `
 <th
   className="clinical-result-time-header-cell padding-standard"
 >
-  <time
+  <div
     className="date"
   >
     December 31st 1999
-  </time>
-  <VisuallyHiddenText
-    text=" "
-  />
-  <time
+  </div>
+  <div
     className="time"
   >
     10:00 PM
-  </time>
+  </div>
 </th>
 `;
 
@@ -84,19 +72,16 @@ exports[`ResultTimeHeaderCell should hide date 1`] = `
 <th
   className="clinical-result-time-header-cell padding-compact"
 >
-  <time
+  <div
     className="date hide-date"
   >
     December 31st 1999
-  </time>
-  <VisuallyHiddenText
-    text=" "
-  />
-  <time
+  </div>
+  <div
     className="time"
   >
     10:00 PM
-  </time>
+  </div>
 </th>
 `;
 
@@ -104,18 +89,15 @@ exports[`ResultTimeHeaderCell should render a date and time 1`] = `
 <th
   className="clinical-result-time-header-cell padding-compact"
 >
-  <time
+  <div
     className="date"
   >
     December 31st 1999
-  </time>
-  <VisuallyHiddenText
-    text=" "
-  />
-  <time
+  </div>
+  <div
     className="time"
   >
     10:00 PM
-  </time>
+  </div>
 </th>
 `;


packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
```diff
@@ -63,7 +62,6 @@ const ResultTimeHeaderCell = (props) => {
       className={timeHeaderCellClassnames}
     >
       <time className={dateClassnames}>{date}</time>
-      <VisuallyHiddenText text={'\u00A0'} />
```
vinaybhargavar, line 66: Does this not work with the space? I think space was added to have a gap between the date and the time.

packages/terra-clinical-result/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
+  * Fixed the spacing between the date in time for the ResultTimeHeaderCell.
```
zxeleven, line 6: nit: `'date and time in the ResultTimeHeaderCell component.'`

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
RayGunY: It's not needed anymore I'm pretty sure. If you look at the screen recordings seems like it's reading out fine to me (I could be wrong)

packages/terra-clinical-result/CHANGELOG.md
RayGunY: good catch https://github.com/cerner/terra-clinical/pull/876/commits/8c82d1fba17437653c3577a8f926b5b8a0176e0e

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
vinaybhargavar: Probably we can confirm with UX on this if the gap is fine.

vinaybhargavar (approved):

RayGunY:
@sdadn @sycombs Because we merged Aaron's changes last week after mine it looks like we missed a piece of validation. His changes didn't get checked against the new terra site test for semantic table usage I created for ResultTimeHeaderCell. Nothing threw a WDIO error because it looks the same in the existing non-semantic examples. And because this sprint I'm creating the main site examples for those semantic tables - we didn't have existing wdio tests to catch this issue.

Anyways, the problem pretty basic, the date and time were squished together when the component is used in a semantic table. This PR has a simple fix for the styling.

@brooksienoodlesoup there isn't much to review here, I'm not sure if you want to/need to take a look.

RayGunY:
Apologies, in JAWs we're noticing the behavior that the visually hidden text was originally added to address. Will probably need to add it back, looking at it now.

sdadn (approved):

packages/terra-clinical-result/src/result-time-header-cell/ResultTimeHeaderCell.jsx
RayGunY: So I noticed after trying to add it back in that it reads out 'space' when using voice over in the semantic table. So when it's there it does fix the JAWs side but it creates an issue on the VO side. We're trying to figure out a solution that will work for both but not sure if one issue is worse than the other from an accessibility standpoint.

vinaybhargavar: I am not sure if it is a good practice but if we use a 'space' (`VisuallyHiddenText text=" "`) directly, VO doesn't read it out. Does using space affect any other way?

RayGunY: To be fair I don't think using visually hidden text like this is best practice anyways. If that works I'm fine with it - did you happen to try that with JAWs as well? I'm having issues spinning up my windows desktop this morning. Maybe @zxeleven can take a look and try on JAWs

zxeleven: @vinaybhargavar I gave that a try yesterday and while it does fix the issue with VO, it reintroduces the JAWS issue of combining the date and time and reading it incorrectly. Going to play around with not using the VisuallyHiddenText component to see if I can find a solution that works for both.

RayGunY: Talked with @brooksienoodlesoup, this looks to be a bigger problem. He said:

> I think the JAWS version sounds worse than the VO version.
> 
> But, I think the problem with this implementation is something other than a needed space.   I think I may have inadvertently given a pass to the HTML validation on how you were using the time element in the examples you coded, when in fact it fails validation.  That's my bad.  We need to use a valid string for date/time in order to validate the HTML code.  To test whether a date time string is valid HTML, use the [Nu HTML Validator](https://validator.w3.org/nu/#textarea), then pick the "text input" method, and type in some HTML code that looks like the date/time string you want to validate, like this:
> 
> <!DOCTYPE html>
> <html lang="">
> <head>
> <title>Test</title>
> </head>
> <body>
> <p><time>Dec 20, 2010 21:00</time></p>
> </body>
> </html>
> 
> This date/time string fails validation when you check it in the Nu validator.  If we are using invalid code, then it is not hard to imagine how a browser/screen reader combo will not read out the contents correctly.
> 
> The HTML5 specification lists valid formats for date and time, which will work with the time element.  You'll want to make sure that the manner in which the date and time is presented in the HTML matches up exactly to one of the formats listed in [the HTML specification for the time element](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-time-element).
> 
> If we can't provide date/time info in a way that is approved by the HTML spec, then I don't think we can expect the screen reader output to work correctly.

Looking at the doc for the time element it looks like **our dates have to be passed in as yyyy-mm-dd**. So in our case we would have to extract the year, month, and day from the date that is passed in and then put it into this format and pass it to the time element instead of the raw date. Our time should be fine as something like hh:mm is standard.

RayGunY: The above solution of pulling out the year, month, and day is going to be involved and idk if we have the time for that. My other thought was to keep the time tag around the "time" but revert the "date" to use divs again. It seemed to read out the date fine without the time tag, it just won't be semantic. At least accessibility is improved for time (which I believe was the one that sounded worse to start) and we could eventually come back to the date portion if needed. Again, it reads out the date pretty reliably even when using a div - I'd have to go re-test through all the browsers in VO and JAWs to ensure that.

RayGunY: After a long conversation with Brooks and Aaron right now we have decided to remove the `<time>` tags entirely and revert back to how this was done before those changes were made. 

This is for several reasons
1. The way our dates are passed in will not work with `<time>` unless we make several changes to how we format those
2. The `<time>` tag is not fully supported by screen readers at this time and may cause differences in how its read between readers and browsers
3. Because JAWs doesn't support time tags it's still reading out the date (and time? I'd have to double check with Aaron on that) in the same way it was before the addition of the time tag. I think the hope was that eventually JAWs would support it, but as of right now the update didn't change much.
4. Voice over didn't particularly benefit from this change, the way it reads out the date and time before and after the time tag change seem to be the same. The difference is there was additional context by adding the time tag but from how it reads out it was already working without the time tag.

With all of the above it seems the cons to using it are larger than the pros, Brooks advises that we should remove the time tags - @scottwilmarth said he trusts what Brooks and we have talked through.

- I tested this branch locally and have added recordings for both the semantic and non-semantic examples. I also tested in edge with VO and it sounds the same. Will have Aaron do some recordings on the JAWs side.


https://github.com/cerner/terra-clinical/assets/40574651/6b8f4c9d-76c5-4833-a271-69b5c4f926c9


https://github.com/cerner/terra-clinical/assets/40574651/a6987eb2-d76e-4c00-806c-a8cfe58b7f34


- Also, I tested what it currently sound like with the time tag in main - can confirm that it sounds the same as when it's not using the time tag.
 

https://github.com/cerner/terra-clinical/assets/40574651/9e53295b-ec4a-42b7-ba98-87ad37abe98d


https://github.com/cerner/terra-clinical/assets/40574651/7ea3ed1e-f9df-4722-8eb2-b1c33f070bcb

RayGunY:
@zxeleven @vinaybhargavar @sdadn @sycombs this pr has changed quite a bit but is good to be looked at again. The basic run down (if you don't want to read my giant comment thread) is that the `<time>` html tag is not working well with the semantic tables/how our dates are formatted. In general it's not supported well by screen readers and since it's bringing up more cons than pros we've decided it's best to just remove it completely.

Talked with Brooks about this and he confirmed we can remove the time tags entirely and revert to how we were handling date and time previously.

brooksienoodlesoup (approved):
Tested in:
Chrome Version 114.0.5735.106 (Official Build) (arm64) + VO (Ventura 13.2.1)
Edge Version 114.0.1823.58 (Official Build) (64-bit) + JAWS 2023.2303.144

Content read out normally with expected behavior.

zxeleven:
Tested with JAWS both with and without the `<time>` tag:

https://github.com/cerner/terra-clinical/assets/37230460/9c79a2b1-cf44-4894-ad9d-e05ed150865e


https://github.com/cerner/terra-clinical/assets/37230460/28b524ee-8f30-4471-be71-1f28a8bfdb58


https://github.com/cerner/terra-clinical/assets/37230460/3c691983-d65c-4b48-898f-87528347af3e


https://github.com/cerner/terra-clinical/assets/37230460/08c6858f-15d8-438c-bef5-94dcf1322932

zxeleven (approved):

vinaybhargavar (approved):

sdadn (approved):
Looks good! @RayGunY thanks for the detailed explanation!

PR #894, opened 2023-08-01
[terra-clinical-onset-picker] Show visible, textual label for all form fields

vinaybhargavar:
### Summary

All form fields must have a visible, textual label, and be programmatically associated with the input it represents to describe the purpose of the form control. Currently, we have labels associated with the inputs that are hidden from the UI using `isLabelHidden`.  We have to unhide/show the labels for all the fields. The recommendation was that the labels should be one word instead of a group of words. Hence, we have to use the placeholder text of each field.

**What was changed:**

Labels are shown on the UI and the placeholder text (single word) of each field is used as a label for each field.

**Why it was changed:**

We need to have labels to associate text with form elements and placeholders alone don't suffice.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

On MAC:
Mac

https://github.com/cerner/terra-clinical/assets/55503680/23ec2607-9e64-4300-91da-8f5931e2b4bc

On Windows:
MS

https://github.com/cerner/terra-clinical/assets/55503680/efc6a83b-f910-469a-9faf-4e6a142d30a4

Deployed Site - https://engineering.cerner.com/terra-clinical/pull/894/

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9245 

---

@cerner/terra


packages/terra-clinical-onset-picker/src/OnsetPicker.jsx
```diff
@@ -451,8 +448,7 @@ class OnsetPicker extends React.Component {
         <SelectField
           className={cx('field-inline', 'age-unit')}
           defaultValue={this.state.ageUnit}
-          isLabelHidden
-          label={intl.formatMessage({ id: 'Terra.onsetPicker.agePrecisionLabel' })}
+          label={intl.formatMessage({ id: 'Terra.onsetPicker.agePrecision' })}
```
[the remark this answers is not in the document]
vinaybhargavar: I think "Age Precision" is fine here because Scott mentioned that labels should generally be one word but not a rule. He approved the changes while doing the initial investigation (POC) for this change but yeah I think we can get his thoughts on this.

@scottwilmarth - Is "Age Precision" fine here to be used as a label or do we need to have a one-word label for this?
Screen Shot 2023-08-04 at 10 49 13 AM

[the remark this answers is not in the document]
scottwilmarth: The actual requirement is not "one word", but instead the label must convey the purpose of the field. Previously, the labels were "Select a Precision," which was not necessary - thus making the labels smaller without the "Select a..." text. "Age Precision" is appropriate in this instance as it communicates the purpose of the selection.

RayGunY (approved):
Looks great Vinay - approved

chenyixuan625 (approved):

vinaybhargavar:
@sdadn, @sycombs - This PR is ready for your review. TIA.

packages/terra-clinical-onset-picker/tests/jest/__snapshots__/OnsetPicker.test.jsx.snap
```diff
@@ -51,7 +51,7 @@ exports[`correctly applies the theme context className 1`] = `
           aria-disabled="false"
           aria-expanded="false"
           aria-haspopup="true"
-          aria-label="Precision, Select a date precision"
+          aria-label="Precision, Precision"
```
sdadn, line 54: Is this aria-label correct?
CC: @scottwilmarth

sdadn (commented):
I'm unsure about the passivity on of switching from always hiding labels to always showing them. Will this break any workflows?

cc: @JessieRandle , @chrismichalewicz , @mjpalazzo

packages/terra-clinical-onset-picker/tests/jest/__snapshots__/OnsetPicker.test.jsx.snap
vinaybhargavar: Yes, one `Precision is from label`, the other is from the placer holder.

vinaybhargavar:
> I'm unsure about the passivity on of switching from always hiding labels to always showing them. Will this break any workflows?
> 
> cc: @JessieRandle , @chrismichalewicz , @mjpalazzo

You are right, the consuming team will have to update their Jest & WDIO snapshots. 

If this needs to be passive, we will need to add a new prop for hiding/showing labels and set the default value to false. 

Thoughts on this? - @scottwilmarth, @sdadn

packages/terra-clinical-onset-picker/tests/jest/__snapshots__/OnsetPicker.test.jsx.snap
scottwilmarth: @sdadn I think this is a perfect example of how we must choose better words when creating props. The aria-label in, this case, is not actual aria-label, but the prop where the first word displays as the placeholder text and the second displays as the visible field label. @vinaybhargavar please correct me if I'm wrong.

Regarding the concern for passivity, these fields will fail accessibility requirements without the faceup label. There is no way to make this a passive change and be accessible. Without the visible label, assistive technology users may not know how to activate the controls - specifically, a speech input device user requires a visible label to know how to call out the controls to activate them.

vinaybhargavar: @scottwilmarth - I don't think that `aria-label` is a prop. We don't set the aria-label here. The clinical onset picker internally utilizes the `SelectField` component from `terra-form-select` which sets the aria-label.

@sdadn - The aria-label remains the same as before, just that the label was changed from Select a date precision to just precision for this field.

vinaybhargavar: @sdadn, @scottwilmarth - To illustrate that the first string is the value of the combo box, and the second string is the label of the combo box, see this line in the snapshot - https://github.com/cerner/terra-clinical/pull/894/files#diff-058379a6a77388b3d64c973d6b9c88743d1472471b2f8ae25d5ad3ed0a3b2b6fR450. First it says `Unknown Selected`, then the label `precision`.

vinaybhargavar: @sycombs - I see that Saad is OOO until Thursday afternoon. Can you take a look at this and let me know your thoughts? TIA.

sycombs (approved):

vinaybhargavar:
@sdadn - Please let me know if you have any other comments here. Thanks.

sdadn (approved):

PR #897, opened 2023-08-08
[terra-clinical-item-view] Correct readout order for two column layout

RayGunY:
### Summary

**What was changed:**
Changed the two column layout to be split by true columns instead of by rows. Now it should read out each column separately top to bottom.

**Why it was changed:**
So that screenreader users can have information conveyed to them the same as visual readers would logically read the information - keeping relevant information close to each other and reading top to bottom.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9429 

**Validation:**

New readout in chrome:

https://github.com/cerner/terra-clinical/assets/40574651/0ea4616e-625b-449e-9033-b947e93ba3ad

New readout in edge:

https://github.com/cerner/terra-clinical/assets/40574651/bf4a6e79-b0a0-4b8e-94aa-f19fee0abcd2


RayGunY:
In this PR several of the 2 column WDIO screenshots are being updated because previously if there was a display with overflow in column 1, then column 2 would have a 'newline' across from that overflow display - even the display on the same line in column 2 didn't have overflow. Now that we've separated by column instead of by row this is no longer happening - I talked with Scott and we think this should be the expected behavior going forward. 

I have two remaining wdio tests failing that I'm working on right now.
One is complaining about the 'o' on this page:
with_accessories

And the other is having some strange content sizing behavior for the enormous snapshot (strange because it's not an issue anywhere else). Looks like maybe the 40/60 split I have between the primary and secondary column isn't working quite the same as the previous 40/60 split of content in the single row. Continuing to investigate, I feel like I'm close but just might need to make some slight adjustments: 
with_two_columns

RayGunY:
I think this is in a good place, WDIO are passing and the only thing that's changed about them is inline with my comment above.

However, one thing I noticed just now that wasn’t caught through WDIO is that when resizing down there is a slight difference in the truncation for two column view. Essentially because these are now split by column instead of row, when resizing/truncating the longest displays start to truncate a little earlier now. Example:

https://github.com/cerner/terra-clinical/assets/40574651/ea5fe1dd-7162-45da-a0bc-9150f53c9419

The overall resizing behavior is technically the same breakdown as before, the only difference is that we can’t encroach into the second column on a row by row basis - so this only effects truncation, overflow is fine. This might help visualize:

Before:
spacing-breakdown-before

After:
spacing-breakdown-now

In the before image we are returning displays by row so the resizing between the two ‘columns’ can be applied on a row by row basis. Essentially we can squish a bit into the extra space of the right side ‘column’ in row three before truncation happens, therefore we don’t truncate as early as we do now with the most recent changes.

With the current changes we are creating true columns in their own containers so the resizing is almost entirely the same, but when truncation is involved we can no longer squeeze column 1's third row display into the extra space of column 2's third row display. They each have their own containers that can’t move into each other.

I'm not sure if this is something we should be concerned with? @scottwilmarth do you have any thoughts?

RayGunY:
Following up in edge on windows, the truncation behavior I mentioned above is also seen there as well. I also noticed some difference in overflow behavior as well. 

New resizing/overflow handling:
chrome-new-overflow

Old resizing/overflow handling:
chrome-previous-overflow

Same reasoning as in my truncation comment applies here. We can't encroach that little bit extra into the right side column so truncation/overflow starts to happen a little early compared to before. This could potentially mess up styling for current consumers - is this something we're willing to have them deal with?

packages/terra-clinical-item-view/tests/wdio/__snapshots__/reference/orion-fusion-theme/en/chrome_medium/item-view-spec/with_word_wrap_in_two_columns.png
sdadn, line 1: Looks like Display 4 & 6 have moved places. Is this an intended change?

RayGunY: Yup, I talked to Scott about it a couple days ago and he said that should be the new expected behavior. One of the comments above was referencing that:

"In this PR several of the 2 column WDIO screenshots are being updated because previously if there was a display with overflow in column 1, then column 2 would have a 'newline' across from that overflow display - even if the display on the same line in column 2 didn't have overflow. Now that we've separated by column instead of by row this is no longer happening - I talked with Scott and we think this should be the expected behavior going forward."

I think especially since we're trying to discourage putting related displays in the two columns row by row since they aren't close in proximity. We'll be adding guidance on that to the documentation page as well as some comments about how the displays are being split between the two columns (since there isn't anything about that on there right now)

scottwilmarth (approved):

RayGunY:
Talked with Scott about the slight resize changes, he said that's ok and gave his +1

As for UX Review for this PR - @chrismichalewicz Krystal mentioned you might be doing those reviews - is that correct or is there someone else I can assign as a reviewer?

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -106,40 +108,49 @@ const defaultEmphasisContentClassesFromIndexes = (rowIndex, rowCount) => {
...
     classes = defaultEmphasisContentClassesFromIndexes(rowIndex, rowCount);
   }
+
   return ['content'].concat(classes);
 };
 
-const renderRow = (row, rowIndex, rowCount, emphasis) => {
-  const rowKey = rowIndex;
+const renderColumn = (displayGroup, displayGroupIndex, emphasis) => {
+  const columnKey = displayGroupIndex;
```
zxeleven, line 139: nit: do we need to assign `columnKey` here or can `displayGroupIndex` be used directly?

RayGunY: We technically could use displayGroupIndex but I've been seeing in this file that typically it's set to a key variable first instead of being directly passed in. I'd probably keep this as is unless we go change the others.

vinaybhargavar (approved):

packages/terra-clinical-item-view/src/ItemView.jsx
zxeleven: Makes sense, thanks for clarifying

zxeleven (approved):

RayGunY:
@sdadn @sycombs this should be good for a final look over whenever you've got a chance

sdadn (approved):

PR #899, opened 2023-08-11
[terra-clinical-item-view] Add prop to override default display styling

RayGunY:
### Summary

**What was changed:**
Created a prop 'overrideDefaultStyling' that defaults to false. When it is set to true the styling for emphasis is skipped and the current styling passed into the display is kept. 

**Why it was changed:**
Adding this prop so that consumers have the option to keep their current styling if necessary. Some display styling is used to convey information such as 'strikethrough' or 'attention'.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9432 

---

chrome-new-example
edge-new-example

https://github.com/cerner/terra-clinical/assets/40574651/75301a61-645d-4453-aead-e26d9039812a


RayGunY:
Site for PR: https://engineering.cerner.com/terra-clinical/pull/899

vinaybhargavar:
@RayGunY - Can we test with JAWS to ensure this works fine?

RayGunY:
> @RayGunY - Can we test with JAWS to ensure this works fine?

@vinaybhargavar sure thing:

https://github.com/cerner/terra-clinical/assets/40574651/21eb987d-1f8a-45df-8c6b-7090017b6793

packages/terra-clinical-item-view/src/ItemView.module.scss
```diff
@@ -62,7 +62,6 @@
 
   .content {
     align-items: flex-start;
-    color: var(--terra-clinical-item-view-content-color, rgba(28, 31, 33, 0.65));
```
supreethmr, line 65: why was this variable removed ? if we are removing the variable then it might have impact on themes which are still using variable `--terra-clinical-item-view-content-color`

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -70,6 +74,7 @@ const propTypes = {
 const defaultProps = {
   layout: Layouts.ONE_COLUMN,
   textEmphasis: TextEmphasisTypes.DEFAULT,
+  overrideDefaultStyling: false,
   isTruncated: false,
```
chenyixuan625, line 78: Just thinking aloud so please correct me if I am wrong. Do we have any potential conflicts between the overrideDefaultStyling and other props(i.e. isTruncated). For example, if the consumer uses their original styling and it is a truncated text, but they keep the isTruncated as false(by default) at the same time. Will this scenario cause any issues?

RayGunY: No it shouldn't cause issues, overrideDefaultStyling just effects the text styling (color, size, bolding, etc) so truncation should not be effected.

RayGunY: Ok so I added some temp examples locally to test some stuff out. Truncated and overflow are working well with the override.

Single column truncation
Screenshot 2023-08-16 at 8 50 11 AM

Two column truncation
Screenshot 2023-08-16 at 8 50 39 AM

Overflow
Screenshot 2023-08-16 at 8 53 04 AM


However, I tested out a display passed in with isTruncated prop as true while our itemview has isTruncated set to false - there were some formatting issues. Left is what it looks like when just itemview has isTruncated set to true, the right is when it's set to false (but the display has it set to true)
Screenshot 2023-08-16 at 8 59 27 AM

This is an existing issue though, with or without the override.

RayGunY: So this will likely need a story logged to fix this use case and for tests/examples to be added for the combo of isTruncated in display as true/isTruncated in itemview as false.

@scottwilmarth tagging you for visibility, should I go ahead and log a jira for the backlog?

RayGunY: Anything related to the above comment will be handled in a different PR, we can just focus on the override logic in this story for now.

RayGunY:
@sdadn @sycombs this should still be good to review whenever you have time
Talked to Scott and he should be taking a look at some point today

scottwilmarth (approved):

vinaybhargavar (approved):

chenyixuan625 (approved):

sycombs (approved):

PR #902, opened 2023-08-15
[terra-clinical-result] Fix useEffect infinite re-render

trandrew1023:
### Summary

**What was changed:**
* Updated the useLayoutEffect back to useEffect
* Removed contentWidth state
* Removed dependencies from useEffect
* Added non-semantic table example (with display:flex on the div wrapping the flowsheetresultcell)
* Updated doc examples to use semantic HTML table
* Added note that when using cell in tables, table styling for column widths should be preserved, or `tableLayout: 'fixed'` can be added

**Why it was changed:**
* Because `display:flex` was being used in a custom GridCell wrapping the flowsheetresultcell, the cell content would render with the same width as the container which would cause an infinite re-render since the container would then change widths upon setNumericOverflow being set to `true`. The widths are compared again the setNumericOverflow would get set to `false` again and so on. To prevent the dependency on the content width, the state was removed and just used directly each time.
* Initially it was thought that using `<td>` outside a table was causing issues, so an example was added with non-semantic table, however it was additional styling. The re-render issue was fixed with the useEffect changes but `width:100%` was added back since it was removed in 1.17.0

### Testing

FlowsheetResultCell with `display:flex` and no `width:100%` before useEffect changes:
image

FlowsheetResultCell with `display:flex` after useEffect changes:
image

FlowsheetResultCell with `display:flex` with `width:100%` after useEffect changes:
image

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9349 

---

@cerner/terra


packages/terra-clinical-result/CHANGELOG.md
```diff
@@ -2,9 +2,16 @@
 
 ## Unreleased
 
+* Added
+  * Added back `width: 100%` to flowsheet result cell styles removed in 1.17.0.
+
+* Changed
+  * Updated FlowsheetResultCell useLayoutEffect to useEffect.
+  * Updated FlowsheetResultCell  dependencies and component states from useEffect.
```
sdadn, line 10: I don't think these changes are relevant or informative to consumers. We can probably change this to:
```
* Fixed:
  * Fixed infinite re-render bug.
```

sycombs: Updated [3ac99ac](https://github.com/cerner/terra-clinical/pull/902/commits/3ac99ac16387f26faca19146515324242f213fc4) and [31b3a3c](https://github.com/cerner/terra-clinical/pull/902/commits/31b3a3c0c7c3e8f24b5b3577489f8ea553bd3e1e)

artpark (approved):

packages/terra-clinical-result/src/flowsheet-result-cell/FlowsheetResultCell.jsx
```diff
@@ -391,27 +391,23 @@ const FlowsheetResultCell = (props) => {
...
-  const [contentWidth, setContentWidth] = useState(null);
   const [numericOverflow, setNumericOverflow] = useState(false);
 
-  useLayoutEffect(() => {
+  useEffect(() => {
     if (!containerDiv.current || !resultDataSet[0]) {
       return;
     }
+    const contentWidth = containerDiv.current.children[0].getBoundingClientRect().width;
+    const containerWidth = containerDiv.current.getBoundingClientRect().width;
```
cm9361, line 401: It does not appear that this logic needed to move outside the if block.  By moving, we would be required to calculate and obtain two bounding client rectangles in cases that are unnecessary.

trandrew1023: Updated in this commit! https://github.com/cerner/terra-clinical/pull/902/commits/7feca5e6dbeee41a971fc1248330b0da352dc0e3#diff-2ef96e32bd2562338959196c9293a261b049ebfc1776e40247b6c8e15014f9b4

sdadn:
Out of curiosity, how was the rerender bug tested?

trandrew1023:
> Out of curiosity, how was the rerender bug tested?

@sdadn It was recreated and tested using one of the consumer projects that had the issue with versions 1.17.0 and above from UXPLATFORM-9349, the code changes were used with the consumer's dev site example locally. An example that causes the infinite rerender and also sometimes the flickering was recreated and added to the terra examples and the new code changes were tested against it to ensure the infinite rerender no longer occurs as well. It was found that external styling required the `width: 100%` that was removed in 1.17.0 which causes the infinite calls the `setNumericOverflow`, so even without the `width: 100%` there is no infinite rerender, however the styling is incorrect, so the width prop was added back (also attached a few screenshots under description showing the different test scenarios and screenshots)

smason0 (approved):

chrismichalewicz:
Functionally reviews, looks good! Updated the label. Great job!

PR #904, opened 2023-08-17
[terra-clinical-label-value-view] Fix padding for dl elements

smason0:
### Summary

**What was changed:**

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra


packages/terra-clinical-label-value-view/src/terra-dev-site/doc/example/LabelValueViewTextTermAndDef.jsx
```diff
@@ -8,6 +8,11 @@ const LabelValueViewText = () => (
         #dl-wrapper > dd {
           margin-bottom: 14px;
         }
+
+        #dl-wrapper {
+          margin-bottom: 0;
+          margin-top: 0;
+        }
```
cm9361, line 15: Why do we need to add a style on the wrapper element if we are updating our sass file?

smason0: This is being applied to the `<dl>` element on line 19 for this specific example and not the one in LableValueView that had its sass file updated. This change is maybe not needed, as it is just an example, but I added it to be consistent in removing these default margins whenever we use a definition list.

nikhitasharma (approved):

sdadn:
How do these changes compare to the original styling utilizing div elements? Are the margins same or different? etc.

smason0:
> How do these changes compare to the original styling utilizing div elements? Are the margins same or different? etc.

The changes make it appear the same as when the wrapper is a div, i.e. the top and bottom margins are 0.

div:
Screenshot 2023-08-24 at 1 29 48 PM

dl:
Screenshot 2023-08-24 at 1 31 28 PM

trandrew1023 (approved):

eawww (approved):

sdadn (approved):

PR #905, opened 2023-08-21
Upgrade to support NPMv9

sdadn:
### Summary

This PR upgrades the repo set up to support NPMv9 so to help upgrade the infrastructure and be up to date with our tech stack. 

Since Node 14 ships with NPMv6, the cicd was updated to utilize NPMv9.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] Other (please describe below)
- [ ] No tests are needed

This was tested by successfully running `npm install` and building the project with both NPM versions (6 & 9).

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

---

@cerner/terra

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -7,6 +7,9 @@ runs:
       uses: actions/setup-node@v3
       with:
         node-version: 14.x
+    - name: Setup NPM v9
+      shell: bash
+      run: npm install -g npm@9
     - name: Clean Environment
       shell: bash
       run: npm run clean
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -10,6 +10,11 @@ runs:
     - name: Setup NPM v9
       shell: bash
       run: npm install -g npm@9
+    - name: Check Node & NPM versions
+      shell: bash
+      run: 
+        - node -v
+        - npm -v
     - name: Clean Environment
       shell: bash
       run: npm run clean
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -10,9 +10,8 @@ runs:
     - name: Setup NPM v9
       shell: bash
       run: npm install -g npm@9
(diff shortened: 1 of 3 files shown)


.npmrc
```diff
@@ -1 +1,2 @@
 package-lock=false
+legacy-peer-deps=true
```
sdadn, line 2: NPMv7+ automatically installs peer dependencies. However, Terra currently has in peer dependency versions which can cause install failures. This flag when set will utilize the legacy behavior of not installing peer dependencies in v6. This will be removed in a future PR when the conflicts are resolved.

sycombs (approved):

package.json
```diff
@@ -4,7 +4,8 @@
   "version": "0.1.0",
   "description": "terra-clinical",
   "engines": {
-    "node": "^10.0.0 || ^12.0.0 || ^14.0.0"
+    "node": "10 || 12 || 14",
```
nikhitasharma, line 7: Does this do an exact version check or supports all minor or patch versions under them. Should this be written like this `^10 || ^12 || ^14`

sdadn: This supports all minor & patch version:
 https://docs.npmjs.com/about-semantic-versioning

I also double checked the syntax using https://semver.npmjs.com.

.github/actions/setup/action.yml
```diff
@@ -7,6 +7,13 @@ runs:
       uses: actions/setup-node@v3
       with:
         node-version: 14.x
+    - name: Setup NPM v9
+      shell: bash
+      run: npm install -g npm@9
+    - name: Check Node & NPM versions
+      shell: bash
+      run: node -v
+      run: npm -v
```
nikhitasharma, line 16: @sdadn  Try this
```suggestion
      run: |
        node -v
        npm -v
```

nikhitasharma (approved):

PR #912, opened 2023-08-30
[terra-clinical-item-view] Add property to separate by row or column when in two column layout

RayGunY:
### Summary

**What was changed:**

- Added a `trueColumn` prop that defaults to true. When it is set to false it reverts to handling two column layout like it was before the changes in this [PR](https://github.com/cerner/terra-clinical/pull/897). This means the screenreader will read out left to right, up to down. Also resizing/truncation/overflow will behave slightly differently (like it was working before) than trueColumn view because displays are being handled by row.
- Added some styling changes to truncated content when it is in true two column view. There was a truncation issue when the ItemView itself didn't have isTruncated set to true but the ItemDisplays passed in did have isTruncated set. Added a check for the displays to see if isTruncated is set on them and if so then to use the truncated two column styling class. Added screenshots below to show the difference. 
- Added several tests for the above behavior of the different potential cases of left, right, or both having truncated displays. 
- I also pulled out some of the logic in the larger methods into their own smaller methods for readability. 

**Why it was changed:**
So that users can have the option to use the old layout/handling for two column in case the new changes break anything for them. Also to fix truncation behavior for partial truncation. 

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9563 

---
VO and JAWs Readouts:

https://github.com/cerner/terra-clinical/assets/40574651/4c30caa5-f941-43f0-8a47-feaf33c7304e

https://github.com/cerner/terra-clinical/assets/40574651/55c5a034-6207-45db-8d1d-15761cdfcb37

True two column resizing behavior is the same as before (truncated and not truncated):

https://github.com/cerner/terra-clinical/assets/40574651/f2ee3018-f2b8-44c4-b0bb-25ab3af8f078

https://github.com/cerner/terra-clinical/assets/40574651/ee82d925-5ccb-4971-bd3c-e00b7cc9cfd1

This is the partial truncation issue I fixed:

Before
partial_truncation_before

After
partial_truncation_after


packages/terra-clinical-item-view/tests/wdio/item-view-spec.js
```diff
@@ -35,10 +48,19 @@ Terra.describeViewports('Clinical Item View', ['tiny', 'small', 'medium', 'large
...
 
 Terra.describeViewports('Clinical Item View', ['enormous'], () => {
-  it('with the full example truncated - one truncated', () => {
+  it('with truncation', () => {
     browser.url('/raw/tests/terra-clinical-item-view/clinical-item-view/overflow-displays-item-view');
 
-    Terra.validates.element('with truncation in one column', { selector: '#ItemView-one-wrap' });
-    Terra.validates.element('with truncation in two columns', { selector: '#ItemView-two-wrap' });
+    Terra.validates.element('with truncation in one column', { selector: '#ItemView-one-truncate' });
+    Terra.validates.element('with truncation in two columns', { selector: '#ItemView-two-truncate' });
```
RayGunY, line 55: Noticed that previously this was not actually pointing to the tests for truncated content. So any of the 'with truncation in one column' or 'with truncation in two columns' snapshots are going to be changed for good reason.

RayGunY: Also some of the 'with word wrap' images have changed a bit as well - just the sizing of that image has changed but not the actual word wrap/overflow behavior. I will post recordings in the description to show this.

packages/terra-clinical-item-view/src/terra-dev-site/doc/clinical-item-view/clinicalItemView.1.doc.mdx
```diff
@@ -37,6 +38,10 @@ import ItemView from 'terra-clinical-item-view';
     title="ItemView - Two Column Layout"
     description="When including icons as part of an '<ItemView.Display />'', it is recommended to use the iconAlignment='inline' prop for best alignment and wrapping with the text."
 />
+<ItemViewTwoColumnByRow
+    title="ItemView - Two Column Layout By Row"
+    description="When the `trueColumn` prop is set to `false` the displays are split two at a time into rows which are stylized to look like columns instead of true columns. So the ItemView is formatted to look like columns but programatically it's handled by row."
```
chenyixuan625, line 43: nit: Same for the `programmatically`

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -216,8 +261,21 @@ const renderView = (displays, layout, emphasis, overrideDefaultStyling) => {
...
+  const displaysSlice = displays.slice(0, 8);
+
+  for (let i = 0; i < displaysSlice.length; i += 1) {
+    if (displaysSlice[i].props.isTruncated === true) {
+      return true;
+    }
+  }
+
+  return false;
+};
```
chenyixuan625, line 274: Looks like this function is specifically used for the truncated fix. Do we want to add some basic JS unit tests on this function?

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -120,26 +125,65 @@ const defaultEmphasisContentClassesFromIndexes = (rowIndex, rowCount) => {
...
+
+const renderRow = (row, rowIndex, rowCount, emphasis) => {
+  const rowKey = rowIndex;
+  return (
+    <div className={cx('row')} key={rowKey}>
+      {row.map((display, displayIndex) => {
+        const displayKey = displayIndex;
+        const contentClasses = classesForContent(rowIndex, rowCount, displayIndex, emphasis);
+        return (
+          <div className={cx(contentClasses)} key={displayKey}>
```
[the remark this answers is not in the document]
RayGunY: We could but I keep seeing places (even in this same file) that have been setting it explicitly as a key then passing it in. I figured I would follow the existing pattern in these methods - unless we go change the others.

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -120,26 +125,65 @@ const defaultEmphasisContentClassesFromIndexes = (rowIndex, rowCount) => {
...
+};
+
+const renderRow = (row, rowIndex, rowCount, emphasis) => {
+  const rowKey = rowIndex;
+  return (
+    <div className={cx('row')} key={rowKey}>
+      {row.map((display, displayIndex) => {
+        const displayKey = displayIndex;
+        const contentClasses = classesForContent(rowIndex, rowCount, displayIndex, emphasis);
+        return (
```
vinaybhargavar, line 177: nit: Can we add a new line before `return` for readability?

RayGunY: [14c6495](https://github.com/cerner/terra-clinical/pull/912/commits/14c6495877fe6eba310beae0211922b7cfbcec50)

packages/terra-clinical-item-view/src/terra-dev-site/doc/clinical-item-view/clinicalItemView.1.doc.mdx
RayGunY: [400fc6d](https://github.com/cerner/terra-clinical/pull/912/commits/400fc6dbe82c75e34d3d09ec6d5ce2e191aa2a2c)

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -26,9 +26,13 @@ const AccessoryAlignments = {
 
 const propTypes = {
   /**
-   * The column layout in which to present the displays. One of `oneColumn`, `twoColumn`.
+   * The visual column layout in which to present the displays. One of `oneColumn`, `twoColumn`.
    */
   layout: PropTypes.oneOf(['oneColumn', 'twoColumns']),
+  /**
+  * Determines whether the displays are programatically separated by row or as true columns when layout is set to `twoColumns`.
```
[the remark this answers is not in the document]
RayGunY: [400fc6d](https://github.com/cerner/terra-clinical/pull/912/commits/400fc6dbe82c75e34d3d09ec6d5ce2e191aa2a2c)

packages/terra-clinical-item-view/src/ItemView.jsx
RayGunY: Ok I went ahead and added a test that should cover this in it: [6fa2823](https://github.com/cerner/terra-clinical/pull/912/commits/6fa2823ad1ab1ee7520979b0f21fd904e32d92b9)

Other related test: [b97a66a](https://github.com/cerner/terra-clinical/pull/912/commits/b97a66a708c12b1e2de898b3296cc26432099523)

chenyixuan625 (approved):

vinaybhargavar (approved):

RayGunY:
@eawww @sycombs @sdadn this is ready to be looked at when you have a chance

eawww (approved):

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -26,9 +26,13 @@ const AccessoryAlignments = {
 
 const propTypes = {
   /**
-   * The column layout in which to present the displays. One of `oneColumn`, `twoColumn`.
+   * The visual column layout in which to present the displays. One of `oneColumn`, `twoColumn`.
    */
   layout: PropTypes.oneOf(['oneColumn', 'twoColumns']),
+  /**
+  * Determines whether the displays are programmatically separated by row or as true columns when layout is set to `twoColumns`.
+  */
```
sycombs, line 34: If the value of `trueColumn` affects screenreader behavior, can we add that to this description?

sdadn:
How does this affect existing behavior? Will this change be passive?

packages/terra-clinical-item-view/src/ItemView.jsx
RayGunY: sure thing: [ec3f4ef](https://github.com/cerner/terra-clinical/pull/912/commits/ec3f4efe61e2a311b363a0ffb3f2e0fd079cd9bc)

RayGunY:
> How does this affect existing behavior? Will this change be passive?

This is essentially adding back old code to make sure we maintain an option for passivity. When this new prop is false it will revert to using the logic we had before the most recent two column accessibility updates.

RayGunY:
> > How does this affect existing behavior? Will this change be passive?
> 
> @sdadn This is essentially adding back old code to make sure we maintain an option for passivity. When this new prop is false it will revert to using the logic we had before the most recent two column accessibility updates.

@sdadn tagging again, just want to make sure you saw my response

sycombs (approved):

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -173,31 +220,32 @@ const renderColumn = (displayGroup, displayGroupIndex, emphasis, overrideDefault
...
-
-      if (count % 2 === 0) {
-        secondaryColumn.push(displaysSlice.splice(0, 1));
-      } else {
-        primaryColumn.push(displaysSlice.splice(0, 1));
+    if (trueColumn) {
+      displayGroups = twoColumnGrouping(displaysSlice);
+    } else {
+      while (displaysSlice.length) {
+        displayGroups.push(displaysSlice.splice(0, 2));
```
sdadn, line 237: Isn't this pushing the same array information to the array multiple times?

packages/terra-clinical-item-view/tests/wdio/item-view-spec.js
```diff
@@ -1,12 +1,19 @@
 Terra.describeViewports('Clinical Item View', ['tiny', 'small', 'medium', 'large', 'huge', 'enormous'], () => {
-  it('with one column displays', () => {
+  it('with displays', () => {
```
sdadn, line 2: I know this is following the previous sentence structure, but since we are updating these, we should reword these so it reads as a sentence starting with "it". E.g. "It renders with with displays" or any other more appropriate descriptor.

sdadn:
> @sdadn This is essentially adding back old code to make sure we maintain an option for passivity. When this new prop is false it will revert to using the logic we had before the most recent two column accessibility updates.

The old logic should be the default for passivity. A few consumers reported the recent updates causing issues. Though the question is if swapping the defaults breaks more or less consumers 😅.

sdadn:
> @sdadn This is essentially adding back old code to make sure we maintain an option for passivity. When this new prop is false it will revert to using the logic we had before the most recent two column accessibility updates.

@RayGunY The old logic should be the default for passivity. A few consumers reported the recent updates causing issues. Though the question is if swapping the defaults breaks more or less consumers 😅.

packages/terra-clinical-item-view/src/ItemView.jsx
RayGunY: No, splice removes the first two displays from displaysSlice and pushes them together to displayGroups. So each time it's taking the top two off and pushing them. It's the same logic that was being used previously.

So displayGroups would look like [[display 1, display 2], [display 3, display 4], [display 5, display 6], [display 7, display 8]]. And by the end displaysSlice would be empty

RayGunY:
> > @sdadn This is essentially adding back old code to make sure we maintain an option for passivity. When this new prop is false it will revert to using the logic we had before the most recent two column accessibility updates.
> 
> @RayGunY The old logic should be the default for passivity. A few consumers reported the recent updates causing issues. Though the question is if swapping the defaults breaks more or less consumers 😅.

I understand what you're saying Saad. @Scott what do you think? I thought we wanted to push people towards using the new logic but have the option to keep the old logic for passivity.

mjpalazzo:
@RayGunY - The conflict between passivity and accessibility is the challenge we are facing. Because consumers have not made their accessibility updates yet, I think we need to go with the passive approach and emphasize in the accessibility guide that consumers must enable the prop when they are remediating for accessibility. Other thoughts?

RayGunY:
> @RayGunY - The conflict between passivity and accessibility is the challenge we are facing. Because consumers have not made their accessibility updates yet, I think we need to go with the passive approach and emphasize in the accessibility guide that consumers must enable the prop when they are remediating for accessibility. Other thoughts?

@mjpalazzo that works for me, just wasn't sure which is the bigger priority in this case. I can go ahead and change the default to use the old logic.

packages/terra-clinical-item-view/tests/wdio/item-view-spec.js
RayGunY: sure thing, [cf417ee](https://github.com/cerner/terra-clinical/pull/912/commits/cf417ee24335fd1281d4babcd523b431e82945ea)

RayGunY:
@sdadn prop default changes have been made here: https://github.com/cerner/terra-clinical/pull/912/commits/456b749ede38e1f2d1741e66ad02ac79b6269292

packages/terra-clinical-item-view/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added `trueColumn` prop so consumers can choose whether to separate displays in the two column layout by row or by column.
```
sdadn, line 6:
```suggestion
  * Added `trueColumn` prop to toggle between displaying in a two column layout by row or by column.
```

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -26,9 +26,16 @@ const AccessoryAlignments = {
...
 const propTypes = {
   /**
-   * The column layout in which to present the displays. One of `oneColumn`, `twoColumn`.
+   * The visual column layout in which to present the displays. One of `oneColumn`, `twoColumn`.
    */
   layout: PropTypes.oneOf(['oneColumn', 'twoColumns']),
+  /**
+  * Determines whether the displays are programmatically separated by row or as true columns when layout is set to `twoColumns`.
+  * Screenreaders will read `trueColumn` displays from top to bottom, one column at a time.
+  * However, this prop defaults to false and therefore the screenreader will read the displays in both columns together, left to right, one row at a time.
```
sdadn, line 35:
```suggestion
  * However, this prop defaults to `false` and therefore the screenreader will read the displays in both columns together, left to right, one row at a time.
```

packages/terra-clinical-item-view/src/ItemView.jsx
```diff
@@ -26,9 +26,16 @@ const AccessoryAlignments = {
...
+  * For accessibility purposes it is recommended to set trueColumn to true.
```
sdadn, line 36:
```suggestion
  * For accessibility purposes it is recommended to set trueColumn to `true`.
```

packages/terra-clinical-item-view/src/ItemView.jsx
sdadn: Ah right, `spice` mutates the original array as opposed to `slice`. +1

packages/terra-clinical-item-view/src/terra-dev-site/doc/example/ItemViewTwoColumnByRow.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import IconBriefcase from 'terra-icon/lib/icon/IconBriefcase';
+import IconPerson from 'terra-icon/lib/icon/IconPerson';
```
sdadn, line 3: We can update these to use the new named exports introduced in https://github.com/cerner/terra-core/pull/3915.

sdadn:
We should probably have an example/code snippet where `trueColumn=false` and maybe compares it with `trueColumn=true` to highlight the differences.

RayGunY:
> We should probably have an example/code snippet where `trueColumn=false` and maybe compares it with `trueColumn=true` to highlight the differences.

@sdadn I believe I already have something that should cover that here: https://github.com/cerner/terra-clinical/pull/912/files#diff-d43cc843d972ecc662a6ef6a0bd194724cf3d97c03658812c7a0f855e209d881R41

The other two column examples on the page are all with trueColumn=true

packages/terra-clinical-item-view/CHANGELOG.md
RayGunY: Fixed: https://github.com/cerner/terra-clinical/pull/912/commits/abe1feec60bf2b34abcc990db0c2964416740638#diff-019a0d835127f70f44b926383f7f9ca20fa8848183637b56a0fe47a22104b394R6

packages/terra-clinical-item-view/src/ItemView.jsx
RayGunY: https://github.com/cerner/terra-clinical/pull/912/commits/abe1feec60bf2b34abcc990db0c2964416740638#diff-a09eb4a886b729c86a83daac740d9a14a92e66fe733a7a907899150b7fdae956R35

packages/terra-clinical-item-view/src/ItemView.jsx
RayGunY: https://github.com/cerner/terra-clinical/pull/912/commits/abe1feec60bf2b34abcc990db0c2964416740638#diff-a09eb4a886b729c86a83daac740d9a14a92e66fe733a7a907899150b7fdae956R36

packages/terra-clinical-item-view/src/terra-dev-site/doc/example/ItemViewTwoColumnByRow.jsx
RayGunY: https://github.com/cerner/terra-clinical/pull/912/commits/abe1feec60bf2b34abcc990db0c2964416740638#diff-8a154576b72d3f91339b6cfa473a6c31078e664975dd661f3f49febe659a5ba4R2

sdadn:
> > We should probably have an example/code snippet where `trueColumn=false` and maybe compares it with `trueColumn=true` to highlight the differences.
> 
> @sdadn I believe I already have something that should cover that here: https://github.com/cerner/terra-clinical/pull/912/files#diff-d43cc843d972ecc662a6ef6a0bd194724cf3d97c03658812c7a0f855e209d881R41
> 
> The other two column examples on the page are all with trueColumn=true

Ah I missed that. Looks good then!

sdadn (approved):

PR #915, opened 2023-09-08
Add Workspaces to project

kenk2:
### Summary

**What was changed:**
Add workspaces to project.

**Why it was changed:**
To keep up with NPM changes.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra


packages/terra-clinical-data-grid/tests/jest/__snapshots__/DataGrid.test.jsx.snap
```diff
@@ -3369,10 +3369,10 @@ exports[`getDerivedStateFromProps correctly applies the theme context className
             className="main"
           >
             <Scroll
-              className="normalizer"
+              className="normalizer orion-fusion-theme"
```
kenk2, line 3372: I am actually not sure why this needs to be updated, but apparently we were never injecting our themes into the datagrid component from lerna before this? Upgrading to lerna v6 apparently requires us to do this.

sdadn: This is not related to lerna, it's due to a dependency generating updated html (likely terra-content-container in this case)

sdadn (approved):

adoroshk (approved):

PR #924, opened 2023-10-06
[terra-clinical-item-view] Fix for null item displays

trandrew1023:
### Summary

**What was changed:**
* Error is thrown when null `displays` prop is passed in, added check (didn't use ? here for readability)
* Error is also thrown when `displays` contains a null item display in the array and a check was added as well

**Why it was changed:**
* Consumers sometimes pass in null to `displays` prop and get an error with the latest changes

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

Before:
image

After:
image

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9711

---

@cerner/terra


adoroshk (approved):

smason0:
Based on the referenced defect, it sounds like the issue seen by the consumer is around `displays` containing null values, rather than the `displays` itself being null (although we should also guard against this). The consuming app breaks when trying to reference the props from the displays array value.

sycombs (approved):

trandrew1023:
> Based on the referenced defect, it sounds like the issue seen by the consumer is around `displays` containing null values, rather than the `displays` itself being null (although we should also guard against this). The consuming app breaks when trying to reference the props from the displays array value.

@smason0  omg really good catch, for some reason I assumed based on the description in the Jira but didn't realize the code snippet was pointing to that. I added the check in this commit https://github.com/cerner/terra-clinical/pull/924/commits/c8619375850a62b4328e70195688680e033b39b1#diff-a09eb4a886b729c86a83daac740d9a14a92e66fe733a7a907899150b7fdae956

smason0 (approved):
+1 with passing build

PR #942, opened 2024-04-11
renamed year select min-width css variable

adoroshk:
### Summary

It has been reported, that the value of the **--terra-clinical-onset-picker-year-select-min-width** variable in fusion theme is can be overridden by the same name variable coming from deprecated **orion-fusion-theme** package, which sometimes appears as dependency in some projects. To avoid that, the variable was renamed to --terra-clinical-onset-picker**-calendar**-year-select-min-width.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

The new variable being populated instead of the old one:
Screenshot 2024-04-11 at 2 25 27 PM

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10364

### DIFF
--- a/packages/terra-clinical-onset-picker/CHANGELOG.md
+++ b/packages/terra-clinical-onset-picker/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Renamed year select min-width css variable for the fusion theme to not be overridden by the same name variable coming from deprecated orion-fusion-theme package.
+
 ## 4.39.0 - (April 2, 2024)
 
 * Changed
--- a/packages/terra-clinical-onset-picker/src/OnsetPicker.module.scss
+++ b/packages/terra-clinical-onset-picker/src/OnsetPicker.module.scss
@@ -46,6 +46,6 @@
   }
 
   .year {
-    min-width: var(--terra-clinical-onset-picker-year-select-min-width, 5.5em);
+    min-width: var(--terra-clinical-onset-picker-calendar-year-select-min-width, 5.5em);
   }
 }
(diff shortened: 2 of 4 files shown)


cm9361:
Just a question, why can't they remove the orion-fusion-theme package dependency?

cm9361 (approved):

adoroshk:
> Just a question, why can't they remove the orion-fusion-theme package dependency

Per request in the UXPLATFORM-10364 that would be change in one place, which is preferable to chasing multiple teams and waiting on their release schedule.

cm9361:
> > Just a question, why can't they remove the orion-fusion-theme package dependency
> 
> Per request in the UXPLATFORM-10364 that would be change in one place, which is preferable to chasing multiple teams and waiting on their release schedule.

Understood.  How do we ensure that there are not other conflicts?  I would want to make sure that we are not making a bunch of one-off variable changes to avoid removing an outdated dependency moving forward.

adoroshk:
> > > Just a question, why can't they remove the orion-fusion-theme package dependency
> > 
> > 
> > Per request in the UXPLATFORM-10364 that would be change in one place, which is preferable to chasing multiple teams and waiting on their release schedule.
> 
> Understood. How do we ensure that there are not other conflicts? I would want to make sure that we are not making a bunch of one-off variable changes to avoid removing an outdated dependency moving forward.

In terra the variable is used in [3 files only](https://github.com/search?q=org%3Acerner%20--terra-clinical-onset-picker-year-select-min-width&type=code). Its has been renamed in all 3 cases. I don't expect any complications in this specific case.

JessieRandle (approved):

